PR #4, opened 2026-07-18
feat!: refactor as a Cobuild-first supervisor

crmapj:
> Merge order: **#5 → #4**. PR #5 establishes the Apache-2.0 and official-branding baseline. After #5 merges, rebase this branch onto `main` and preserve that metadata. CI and release work is handled separately in #23.

## Summary

This PR turns `dku-headless` into a purpose-built supervisor layer for agents working with Dataiku DSS. The external agent gathers context, delegates in-project construction to Cobuild, and verifies the result independently. It can directly execute existing assets, but it does not maintain a second set of general-purpose write tools beside Cobuild.

That boundary is the core design choice: one system builds, another checks. It keeps authority clear, reduces accidental mutation, and makes the exposed MCP contract small enough to test and review as a whole.

### One fixed MCP surface

The server now exposes one curated contract: **53 non-Cobuild tools plus 5 Cobuild tools**. The old exposure modes, `DKU_MCP_COBUILD_MODE`, `DKU_MCP_TOOL_EXPOSURE`, and the BM25 search transform are removed.

- Read and inspection tools ground the agent in the current DSS state.
- General in-project create, update, and delete tools are removed; those changes go through Cobuild.
- Five bootstrap endpoints remain for work that must happen before Cobuild can act, or that transfers data from the agent's environment.
- `build_datasets`, `run_recipe`, and `run_scenario` are the only direct execution tools, because they run assets that already exist rather than design new ones.
- Transport gating remains explicit. Streamable HTTP removes tools that assume access to the server's filesystem or shared mutable instance selection, and provides row-based upload instead.

The exact tool sets for both stdio and streamable HTTP are pinned in tests. This makes the surface a versioned contract rather than an incidental result of imports or environment flags.

### Durable Cobuild delegation

The five Cobuild tools implement a conversation and turn lifecycle rather than a single blocking request:

- Conversations are stored under `DKU_MCP_STATE_DIR` with locked, atomic, permission-restricted writes and are rehydrated after a server restart.
- `allow_edit_project` is a per-message grant and defaults to `false`, so inspection and planning remain read-only unless the caller grants edit authority for that turn.
- Long turns continue after the client timeout. The caller receives a stable turn handle and polls `get_cobuild_turn_status` instead of sending the instruction twice.
- In-flight overlap guards and a cross-process concurrency cap prevent duplicate or unbounded work when several MCP processes share the same state directory.
- Destructive proposals return a `confirmation_id`. Approval must echo that exact identifier, binding consent to the deletion the agent inspected.
- Ambiguous transport failures, lost turns after restart, abandoned work, and saturation have distinct outcomes with recovery guidance. Unknown mutation outcomes are never treated as safe to retry blindly.

This design accepts that network requests and build turns have different lifetimes. Durable metadata preserves continuity, while explicit unknown-outcome states prevent a timeout or restart from becoming a duplicate mutation.

### Composed grounding and independent verification

Three higher-level reads reduce the number of calls needed to understand and finish a project:

- `get_project_overview` gathers project identity, flow objects, recent activity, variables, and other orientation data with per-section failure isolation.
- `get_flow_graph` returns nodes, edges, and a bounded dependency tree, so the agent can reason about structure rather than infer it from flat lists.
- `audit_project` is a read-only finish gate over structure, documentation, evidence, and maintainability. Optional contracts can assert expected columns, types, row counts, and non-blank outputs.

The audit path fails closed: unavailable or unreadable evidence cannot become a pass. Sampling, metrics, traversal, logs, and library reads are bounded so verification cannot silently turn into an unbounded workload. Secret-like project variables and connection values are redacted before they reach the model.

### Bounded execution of existing assets

`build_datasets`, `run_recipe`, and `run_scenario` restore the useful execution paths without reopening general write access. They return real job or scenario-run identifiers, avoid unbounded SDK waits, and let the supervisor inspect completion through the read tools. `run_recipe` follows the flow through its output so schema updates use DSS's normal build behavior.

Separating execution from construction lets an agent run and verify an approved design without giving it a parallel API for changing that design.

### One skill, with generated references

The 23 per-object skills are replaced by one `dataiku-headless` skill:

- a compact `SKILL.md` router for permanent rules and task routing;
- `soul.md` for supervisory judgment and decomposition;
- five playbooks for building, inspection, direct execution, verification, and migration;
- focused references that are loaded only when needed;
- a generated tool index derived from the live FastMCP registry.

CI checks that every routed file is reachable, every path resolves, every documented tool name exists, and the generated index matches the registered server. Keeping each rule in one place cuts duplicated guidance while the checks stop the skill and server from drifting apart.

### Configuration, authentication, and runtime safety

Configuration now has one documented resolution path: `DKU_CONFIG_DIR`, then the repository-local `.dataiku` directory, then the user config directory. Request bearer authentication survives worker-thread boundaries, tool calls bind their instance and client at entry, and strict parsing rejects unclear TLS settings. Streamable HTTP never exposes a tool that reads an arbitrary server-local path.

## Breaking changes

- Removes the old tool-exposure and search modes.
- Removes broad direct CRUD surfaces for in-project objects.
- Replaces the per-object skill collection with the single routed skill.
- Makes `confirmation_id` mandatory when answering a Cobuild deletion request.
- Makes Cobuild editing opt-in for each message.
- Removes obsolete skill-install and initialization paths now that plugins and repository-native MCP configuration are the distribution mechanism.

## Verification

- **207 tests pass** on this branch.
- Ruff passes.
- Skill link, routing, documented-tool-name, generated-index, and exact tool-surface checks pass.
- CI runs the locked environment, lint, tests, and skill checks on Python 3.10 and 3.14.


pmasiphelps (commented):
I have a few high-level comments:

1. This is a big refactor. Can we take this idea of creating the single new dataiku-headless parent skill and then reference the object-specific skill files we had before (and make them ref files instead)? 

2. We're moving away from the repo's previous implementation of inspection tools (get_, list_ on many objects), and instead asking Cobuild for this info, referencing the object-model.md file. Why this change? Have you tested this out?  If so, how does it seem?

3. The new wording in the dataiku-headless moves the language more aggressively toward headless being a supervisor of Cobuild, rather than the previous language ... 

> An MCP server and agent skill library for operating Dataiku with an AI agent harness (Claude Code, Codex, Snowflake CoCo (Cortex Code), Cursor, OpenCode, or a custom agent). Connect your agent to a Dataiku instance to inspect projects, gather context, and drive Cobuild, Dataiku's AI building agent to build data pipelines, analytics, machine learning models, multi-agent workflows, applications, and automation pipelines inside Dataiku. Cobuild is exposed here as a retained conversation, driven through MCP tools. This repo's own tool surface stays deliberately thin around it: read/list/get/inspect tools for every object type (for context-gathering inside or outside a Cobuild conversation), plus a handful of operations Cobuild cannot do because they are cross-project, instance-level, or precede a project/conversation existing (creating a project, uploading a local file into a dataset or managed folder or project library).

We'll want to add the admin CRUD tools in, so I think the new language is a bit too aggressive as-is.

4. soul.md reads like AI slop. Some of the concepts in here are good, and previously, we covered these types of things in the skills docs - in a "concepts", "workflow", or "safety rules" section. Why move these to soul?

crmapj:
Thanks Pat, really good points, all four. The changes are on this PR now, refs below.

**Single parent skill + old files as references.** That's basically what this is, with one catch: the old files couldn't survive as refs, since most of their content documented the deleted tool surface and repeated the same routing/safety boilerplate across 23 skills. Measured: roughly 1,700 hand-written lines went to 580, about two thirds of that being duplication and dead tool docs, with the judgment content moving over nearly 1:1. The shape follows [Anthropic's skill spec](https://www.anthropic.com/engineering/equipping-agents-for-the-real-world-with-agent-skills) (router plus on-demand references; "many narrow, single-use skills" is a named anti-pattern there) and comes straight from dku-cli, where we ran these layouts against real build tasks and saw the structure work well, including having a soul as the judgment layer. Once this merges we can test and benchmark this repo's profile the same way. Three real losses did slip through, restored in [`3e0908d`](https://github.com/dataiku/dku-headless/commit/3e0908d): the formula language doc, the prepare processors doc (we still read recipe settings, so we need both) and the migrations contributing guide. And if per-object contributability is the concern, splitting `object-model.md` into `references/objects/<type>.md` under the same router is easy, happy to do that.

**The inspection tools.** The pruning follows [Anthropic's tool guidance](https://www.anthropic.com/engineering/writing-tools-for-agents) (consolidate narrow tools into parameterized ones; bloated toolsets are "one of the most common failure modes"), and GitHub's MCP server recently made the [same cut for the same reason](https://github.com/github/github-mcp-server/discussions/1182), 101 down to 52 default tools. But you caught a real problem: for agents, dashboards, wikis and semantic models the only read path left was asking Cobuild, i.e. the builder grading its own homework. Fixed in [`9edf6fd`](https://github.com/dataiku/dku-headless/commit/9edf6fd): one generic `get_object_settings(project_key, object_type, object_id)` over a 13-type enum, using the same `dataikuapi` calls the old tools did, secret-redacted and size-bounded. It's been live-tested read-only against a real instance and then hardened through an adversarial model review in [`610787a`](https://github.com/dataiku/dku-headless/commit/610787a).

**The aggressive language.** Agreed, softened in [`6e184cc`](https://github.com/dataiku/dku-headless/commit/6e184cc). The routing rule is now stated by capability rather than counts: in-project asset work goes through Cobuild; direct tools exist where Cobuild can't act (instance-level, cross-project, bootstrap, deterministic execution of existing assets). Admin CRUD is instance-level, so it lands inside that rubric without rewriting anything. I didn't build any of it, that's yours.

**soul.md.** Fair hit on a few lines, tightened in [`63b26e1`](https://github.com/dataiku/dku-headless/commit/63b26e1) (the numbered "doctrine" cross-refs and a couple of flourishes are gone). The pattern itself is inherited rather than generated though: it's adapted from dku-cli's `soul.md`, and the old per-object safety rules didn't move there, they live in SKILL.md's permanent rules and `safety-and-confirmations.md`. What soul.md holds is the cross-cutting supervision judgment (decompose before delegating, stage gates, verification stances) that had no single home across 23 files and would otherwise get duplicated into every playbook.

nmartorell:
Before going into the review, I would first like to say that I truly appreciate the contribution. I can see many ways in which the ideas in this PR will make the dku-headless project materially better. That being said, I concur with @pmasiphelps' review, but would like to go slightly further: I do not think this PR is mergable, in its current or any future state. 

The reasons for this are:

1. Scope: there are far too many functional, structural and interface changes included into a single PR
2. Peer review: the number of changes and interdependence between them make this PR is objectively unreviewable
3. Author review of AI-generated code: there are strong indications that the code and skill files were not reviewed by the author(s) prior to submission


**Scope**
I found it hard to keep track all the different changes the PR attempts to make. Most of them are strongly opinionated architectural decisions, some of which I agree with, some of which I do not. But this is slightly beside the point: the number of changes makes it impossible to have an productive discussion on the tradeoffs of each proposed change. A discussion on this PR would become instantly unwieldy.

To illustrate this point, here is a list of just a handful of the proposed changes (with some notes):

-  Skills: this is not just a multi-to-mono skills refactor, but a complete change in skills architecture, going from a dataiku-object-centric to objective-centric design. In addition, the "sentiment" of the skills also changes, moving from a mostly agnostic set of instructions to heavily opinionated ones (e.g. on the use of flow zones, descriptions, etc.). 
- Addition of soul.md
- Removal of default connections and LLM models in the configuration
- Removal of MCP search mode
- Additional of persistent Cobuild conversation history
- Removal of many "list/get" tools (and MCP "Cobuild mode")
- The addition of "mega" and/or opinionated tools
- ... and probably a bunch more that I missed

What I want to highlight here is the sheer number and scope of changes, and less my agreement/disagreement with any change in particular. I am more than open to discussing any and all of the proposed changes in a more orderly fashion. 


**Peer review**
Properly reviewing over 12k additions and 4.5k deletions is... just not going to happen. At least not in one go. It would take us days to do so in any meaningful way. The authors and I likely have strong philosophical disagreements as to the appropriate use of AI for code-generation and review; I think it is important that we discuss this.

More importantly, the large number of changes make it impossible to measure the effect each change has on the behavior and output of dku-headless. I can't see how we could possibly make an informed decision on which feature(s) to keep without having a quantitative way of measuring their effect. On this point, a reminder: the current state of dku-headless passes all bechmark tests, so we'd be looking to make incremental cost and execution time improvements.


**Author review of AI-generated code and skills**
I think it's reasonable to assert that, given the large numbers of changes (measured corsely in lines of code/skills), and the time-frame in which these changes have been proposed, it is unlikely that the proposed PR has been thoroughly reviewed prior to submission. Conjectures aside, there are several skills documents that present logical inconsistencies that would be hard for a human reviewer to miss (as an example, line 48 of SKILL.md seems to be leaked internal LLM reasoning; the migrations skill refers to different md docs in different steps, etc.). 

This probably comes back to the philosophical question mentioned above, which I strongly feel we must discuss; namely, the level of responsibility we have, as humans, to review LLM-generated code prior. Personally, I do not believe unsupervised LLM/agents are as of-yet capable of creating maintainable high-quality codebases (and in fact require high amounts of review and supervision).


**tl;dr**
To summarize: I don't think this PR can be salvaged in its current state. It proposes too many non-orthogonal changes, imposing an undue burden on the reviewers (who have almost no other option than to blindly accept it, propose near-cosmetic edits, or reject it outright).

I propose re-grouping tomorrow to (at the very least) agree on the following:
- Operations: what should the extent of a PR be? How do we decide on if a PR should be merged?
- AI: what is our responsibility as engineers, colleagues, and most importantly humans, when working with AI? What is our purpose?
- Roadmap (to next Friday and beyond): let's actually decide what we **should** develop, and not just develop because we **can**.

crmapj:
Happy to split it up into multiple PRs. The commits are pretty atomic so found it easiest to review that way. All changes have come from the action items discussed last week.

PR #20, opened 2026-07-20
docs(skill): Excel-only migration playbook; tighten Cobuild turn guidance

pportal2:
Adds a self-contained `playbooks/excel-migration.md` and makes Excel the skill's only migration path: the generic migration playbook and the `references/migration-sources/` tree (Alteryx/SAS references, helper scripts, contributing guide) are removed, with Excel migrations routed through the shared Cobuild supervision and verification rules. The Cobuild turn guidance is also tightened — status routing kept at the decision points, verification scoped from Cobuild's report, fail-closed on empty deletion scope — while duplicated schema fields, repeated examples, and unsupported timing claims are dropped. The guidance is grounded in real Excel and Alteryx migration runs supervised through this surface, most recently a live Excel project-finance migration (Opus 4.8) that exposed three gaps now fixed in the playbooks: build contracts must pin a `test_connection`-verified storage connection per output (Cobuild otherwise free-picks, and picked a broken Snowflake connection), Group/Pivot units must spell out output column order, and a task is not done while a Cobuild turn is still in flight — poll until it settles and only then run the `audit_project` gate. Verified with the skill link checker, the tool-name checker, and the full test suite (244 passed).


yjouini:
please rebase @pportal2

pmasiphelps:
@pportal2 can you please rebase to main?

PR #21, opened 2026-07-20
docs(readme): capability-based direct-tool rubric, room for instance admin

crmapj:
## Reviewer concern

pmasiphelps flagged on PR 4 that the README was too aggressive about headless being purely a Cobuild supervisor. The team plans to add instance-admin CRUD tools (users, groups, connections, code envs) later, and the hard enumerations — "Three direct executions … Nothing more runs outside Cobuild", "Four bootstrap writes", "The only writes Cobuild cannot do" — would become lies the day those tools land.

## The fix: capability rubric, not counts

The routing rule is now stated by capability:

> in-project asset work routes through Cobuild; direct tools exist where Cobuild cannot act — instance-level operations, cross-project operations, bootstrap actions that precede a project or conversation, and deterministic execution of assets that already exist.

Plus one forward-looking sentence: the direct surface grows only along those capability lines (e.g. instance administration) — never with in-project asset writes.

## Why counts moved from rule to illustration

Counts as the rule turn every tool addition into a doc contradiction. Counts as illustration ("today that's `build_datasets`, `run_recipe`, `run_scenario`") stay accurate as a snapshot and let instance-admin tools land inside the stated principle without rewriting the doc. The supervisor thesis is untouched: Cobuild remains the only path for in-project asset mutation, and the "no direct `create_recipe` / `update_dataset`" line is unchanged.

## Out of scope

- No admin CRUD tools are added or documented here — they don't exist yet and another dev owns them.
- No files other than README.md are touched.


crmapj:
Folded into PR #4 directly as 6e184cc (per review discussion there).

PR #28, opened 2026-07-21
feat(cobuild): retain and poll long-running turns

crmapj:
# Why this is worth doing

Cobuild work can take minutes. If the MCP client stops waiting, that does **not** prove the DSS-side action failed. Sending the same mutation again can duplicate work or race the first request.

This change gives every Cobuild turn a pollable ID, keeps the original worker authoritative after a timeout or client disconnect, and blocks overlapping work until the first turn reaches a durable result.

This is draft **part 4 of 13** in the PR #4 review stack. It depends on #27, which supplies the private, owner-bound conversation store. Keeping this slice separate lets reviewers focus on one question: **does long-running Cobuild work have a safe lifecycle?**

## What changes

- Run each blocking Cobuild turn in its own named daemon thread, outside the shared short-call executor.
- Mark its `Future` as running before exposing it to an async wait, so cancelling the MCP wait cannot cancel a worker that has already been accepted.
- Return `status=timeout` plus a `turn_id` when the wait budget expires; the worker continues and can be polled through `get_cobuild_turn_status`.
- Reject overlap per conversation in both the current process and other MCP processes sharing the state directory.
- Enforce a bounded cross-process turn capacity with private Portalocker slot files.
- Persist the exact terminal result before releasing the conversation and capacity locks.
- Distinguish a live worker in another process (lock held) from a worker lost in a process exit (durable `in_flight` marker, lock free).
- Keep overdue workers locked and visible. This code never guesses that a thread is dead, releases its lock, and admits a competing mutation.
- Preserve the complete deletion proposal before making its confirmation ID usable. A failed final write keeps the turn locked and hides the ID until a later poll succeeds.
- Bound retained handles, settled turns, environment settings, and persisted terminal responses.
- Document the poll flow and its operator settings in the skill, README, and `.env.example`.

## Safety contract

| Invariant | Enforcement | Test evidence |
| --- | --- | --- |
| One active turn per conversation | Process registry plus a hashed cross-process conversation lock | Two concurrent sends produce one SDK call; a real child process owns the external-lock case |
| A timeout is not a cancellation | Running `Future` is retained; timeout only ends `asyncio.wait` | Timed-out worker completes once and returns the same `turn_id` on poll |
| A disconnected client cannot cancel accepted work | `Future.set_running_or_notify_cancel()` runs before the worker is exposed | Cancelling the waiting task leaves the worker alive and pollable |
| Capacity applies across processes | One Portalocker slot is held for the full turn | A child process holding the only slot saturates the parent |
| Terminal state cannot lag behind unlock | Store write happens before both OS locks are released | Injected disk failure blocks a successor; polling retries finalization |
| A live peer is not called “lost” | Status probes the conversation lock | Held external lock returns `in_progress`; only the released orphan becomes `turn_lost` |
| A hung thread cannot overlap a successor | Running turns are never swept or unlocked; only settled turns are evicted | Artificially overdue turn remains registered, locked, and exclusive |
| Delete consent always matches visible evidence | Full proposal and exact ID are one durable terminal result | Approval is blocked during persistence failure; oversized proposals are disarmed, not truncated |
| Results and registries are bounded | 512 KB terminal ceiling, 256 live handles, 64 settled turns, configured capacity ceiling | Large-proposal and bounded-registry paths are covered |
| Ownership does not drift during an await | Instance, client, principal, and store are captured before handoff | Existing credential/instance/project tests plus retained-store capture regression coverage |

### The turn ID is a poll handle, not an idempotency claim

[Google AIP-151](https://google.aip.dev/151) recommends returning a token for a long-running operation so callers can check back instead of blocking. That is the model used here.

[Google AIP-155](https://google.aip.dev/155) is stricter: a request ID may imply idempotent retries only when the server can guarantee de-duplication. The public Cobuild call used here does not accept such an idempotency key. Therefore this PR does **not** pretend that a blind resend is safe. The `turn_id` identifies the accepted local operation for polling; ambiguous mutations remain “inspect before retry.”

### A dedicated daemon thread avoids starving ordinary MCP reads

The [Python 3.10 Future documentation](https://docs.python.org/3.10/library/concurrent.futures.html#future-objects) defines the running/cancellation and completion-callback rules this state machine relies on. The same documentation warns that `ThreadPoolExecutor` workers are joined at interpreter exit and are not recommended for long-running tasks. A dedicated bounded daemon thread keeps a multi-minute Cobuild call out of the small executor used by ordinary DSS reads.

A daemon thread still does not survive a process exit. The durable marker plus OS lock makes that limit explicit: a held lock proves a live peer; a free lock beside `in_flight` means the outcome is unknown.

### Locks are cross-platform and represent live ownership

[Portalocker](https://portalocker.readthedocs.io/en/latest/) provides cross-platform file locks, and its [Lock/BoundedSemaphore API](https://portalocker.readthedocs.io/en/latest/portalocker.html) documents the same lock and process-wide capacity patterns used here. All server processes that share `DKU_MCP_STATE_DIR` must use the same configured capacity.

The lock file is only a rendezvous point; the OS lock is the ownership signal. Deleting a lock file cannot safely stop a worker, so the returned overdue guidance explicitly says not to do that.

### Deletion approval waits for durable evidence

A terminal result containing `needs_confirmation` is written with its full `objects_to_delete`, `deletion_impacts`, and exact `confirmation_id` before the turn unlocks. If that write fails, the proposal is reported only as `finalization_pending`; no approval token is exposed. Polling retries the same finalization while every successor remains blocked.

This keeps the least-privilege and transaction-specific consent rules established in #26 and the private durable storage rules established in #27.

## Dataiku and original-PR provenance

- Dataiku’s [official project API reference](https://developer.dataiku.com/latest/api-reference/python/projects.html#dataikuapi.dss.project.DSSProject.new_cobuild_conversation) documents `new_cobuild_conversation()`, the synchronous `send_message(..., allow_edit_project=...)` call, and its permission behavior. This adapter adds lifecycle supervision around that public API; it does not replace the DSS operation.
- The retained-turn idea originated in PR #4’s [delegation-core commit](https://github.com/dataiku/dku-headless/commit/b66800ba8112f009716d46b49d60ba05515692f3), with later hardening in [085abac](https://github.com/dataiku/dku-headless/commit/085abacffdfed510a36e26a815d64eddd7144c7c), [465be40](https://github.com/dataiku/dku-headless/commit/465be40d13f0e04fe8fb768f094335af75f8a917), [d78e952](https://github.com/dataiku/dku-headless/commit/d78e952276fd406eb5698c3d7b12741b66b0ea5e), [87d7fa0](https://github.com/dataiku/dku-headless/commit/87d7fa0fa91b99cc78fcf1d2aac703b752d1173e), and [4f26971](https://github.com/dataiku/dku-headless/commit/4f2697174e080d76015b1c26c4d47a3266a50b9c).
- The split answers the scope concern in the [PR #4 review](https://github.com/dataiku/dku-headless/pull/4#pullrequestreview-4737994324) and the request to make the work reviewable in [this comment](https://github.com/dataiku/dku-headless/pull/4#issuecomment-5028205473).
- This slice is **not** copied from `dku-cli`. The CLI is useful provenance for later audit and skill slices; retained MCP request ownership is adapter-runtime behavior.

This version deliberately corrects three unsafe edges found while fully reviewing PR #4:

1. it uses Portalocker instead of POSIX-only `fcntl`;
2. it never frees a still-running “hung” worker’s locks; and
3. it probes the OS lock before declaring a durable `in_flight` marker lost, avoiding a false `turn_lost` report for work still running in another process.

## Scope

Included:

- retained send and confirmation turns;
- polling and outcome recovery;
- cross-process overlap/capacity claims;
- terminal-result and complete-proposal durability;
- operator and skill guidance for this lifecycle.

Not included:

- direct dataset/recipe/scenario execution (next slice);
- the composed context and audit tools;
- final tool-surface pruning and the one-skill rewrite;
- server-side retry de-duplication, which the public Cobuild API does not offer here.

## Verification

Run on both supported interpreter lines:

```text
uv run --python 3.10 ruff check .     # passed
uv run --python 3.10 pytest -q        # 90 passed
uv run --python 3.13 ruff check .     # passed
uv run --python 3.13 pytest -q        # 90 passed
uv run cz check --rev-range origin/agent/pr4-03-durable-conversations..HEAD
                                        # successful
```

The suite includes an actual subprocess, not only mocks, for external ownership, lost-turn detection, and the global capacity slot.

## Suggested review order

1. `ConversationStore.claim_turn`, `record_turn_result`, and `inspect_turn`
2. `_begin_turn`, `_wait_for_turn`, and `_try_finalize`
3. the three public send/answer/status paths
4. the retained-turn, subprocess, cancellation, confirmation, and bound tests
5. the operator and skill wording

## Stack

- #25 — deterministic runtime configuration
- #26 — least-privilege Cobuild requests and exact deletion consent
- #27 — private, owner-bound durable conversation state
- **this PR — retained turns and polling**

Draft only: please review this slice before it is marked ready.


crmapj:
Closed in the rework: retained turns and the conversation lifecycle became one design, now in #27. The durable store was dropped for v1; the retained-turn half lives on there.

PR #33, opened 2026-07-21
refactor(mcp)!: one fixed tool catalog, stdio only

crmapj:
Rework of the fixed catalog PR, now also removing the HTTP transport.

What this PR does:
- Deletes both tool-visibility systems: the BM25 search mode and DKU_MCP_COBUILD_MODE. What is registered is what you get, on every install.
- Removes the streamable HTTP transport entirely. The server is stdio only.
- One test pins the exact registered tool set as the surface contract.

Why stdio only: with modes or per-transport differences, the same agent instruction works on one deployment and fails on another, which is impossible to support. HTTP also demands credential ownership and shared-state machinery a local single-user plugin should not carry. If central hosting becomes a need, it returns as its own project.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

----
This PR addresses these Trello cards from the backlog:
- Remove streamable-http mode: https://trello.com/c/v8MbiEvM/145-strip-out-streamable-http-mcp-server-model
- Remove MCP Search mode: https://trello.com/c/feEISziI/135-remove-mcp-tool-search-mode
- Remove Cobuild mode: https://trello.com/c/GYPCoypy/149-remove-mcp-cobuild-mode

PR #34, opened 2026-07-21
refactor(tools)!: prune only proven duplicate readers

crmapj:
## Why this change

This PR removes 16 duplicate settings readers, but keeps every discovery,
runtime, search, and verification tool that returns distinct evidence.

That is the reviewable boundary: fewer overlapping choices for an agent without
making Cobuild grade its own work. It also hardens the retained project-library
tools so large trees, files, responses, or hostile regular expressions have
explicit limits.

## What changes

- Replace 15 per-object settings/version readers with the already-reviewed
  `get_object_settings` dispatcher.
- Replace `get_dataset_column_descriptions` with a bounded
  `get_dataset_info(columns=[...])` response that retains caller-ordered
  selection and comments while adding type, meaning, and connection.
- Keep 25 tools that the original PR #4 pruning removed even though they expose
  distinct discovery or runtime evidence.
- Pin the resulting contract at **79 non-Cobuild + 5 Cobuild tools** before the
  transport gate.
- Bound project-library traversal, reads, search, validation, and final serialized
  responses. User regexes use real match timeouts.
- Record every remove/keep decision in
  [TOOL_SURFACE_DECISIONS.md](https://github.com/dataiku/dku-headless/blob/agent/pr4-10-pruned-surface/TOOL_SURFACE_DECISIONS.md).

## The simple decision rule

Remove a narrow reader only when another tool returns the same evidence or a
strict superset. Keep a tool when it discovers an id, observes live state,
searches content, or verifies an outcome that settings cannot show.

This follows Anthropic's advice to consolidate overlapping tools, while avoiding
the failure identified in the [first PR #4 review](https://github.com/dataiku/dku-headless/pull/4#pullrequestreview-4737994324):
asking the builder what it built is not independent inspection.

## Removed tools and exact replacements

| Removed | Replacement |
|---|---|
| `get_agent_review` | `get_object_settings(object_type="agent_review")` |
| `get_agent_settings`, `list_agent_versions` | `get_object_settings(object_type="agent", version_id=...)` |
| `get_agent_tool_settings` | `get_object_settings(object_type="agent_tool")` |
| `get_dashboard_settings` | `get_object_settings(object_type="dashboard")` |
| `get_dataset_column_descriptions` | `get_dataset_info(columns=[...])` |
| `get_evaluation_store_details` | `get_object_settings(object_type="evaluation_store")` |
| `get_insight_settings` | `get_object_settings(object_type="insight")` |
| `get_knowledge_bank_settings` | `get_object_settings(object_type="knowledge_bank")` |
| `get_ml_analysis_settings` | `get_object_settings(object_type="ml_analysis")` |
| `get_retrieval_augmented_llm_settings` | `get_object_settings(object_type="retrieval_augmented_llm")` |
| `list_saved_model_versions`, `get_saved_model_version_details` | `get_object_settings(object_type="saved_model", version_id=...)` |
| `get_semantic_model_version_settings` | `get_object_settings(object_type="semantic_model", version_id=...)` |
| `get_webapp_settings` | `get_object_settings(object_type="webapp")` |
| `get_wiki_article` | `get_object_settings(object_type="wiki_article")` |

The generic inspector calls the same `dataikuapi` accessors, recursively
redacts secrets, discovers versions when `version_id` is omitted, and enforces
a hard 1 MB final response ceiling. Its dispatch and failure cases have dedicated
tests for all 13 object types.

## What deliberately stays

The original surface commit, [`c753f9b`](https://github.com/dataiku/dku-headless/commit/c753f9b75e32bf59521b8e46c5df23e2c7d9f04d),
cut further than equivalence supports. This PR retains:

- object discovery for reviews, tools, dashboards, evaluations, insights,
  knowledge banks, RAG LLMs, semantic models, WebApps, and wikis;
- Agent Review tests, runs, and results;
- live WebApp backend state;
- individual Data Quality rules, histories, and results;
- LLM detail and Knowledge Bank retrieval;
- trained-model discovery, summaries, metrics, and detail;
- messaging-channel discovery needed by scenarios;
- project-library search and Python validation.

These are not alternate spellings for settings. They answer different questions.

## Project-library hardening

The retained library tools now have independent limits for:

- returned items and all visited nodes;
- tree depth;
- attempted file reads;
- accepted bytes per file;
- returned matches;
- Python validation input;
- per-regex match time;
- cooperative traversal/matching time;
- final serialized response size.

Partial results carry `truncated: true` and named reasons. The visited-node
counter includes nodes excluded by `source`, fixing the original implementation
where a selective filter could walk unlimited non-matches. Traversal is iterative
to avoid recursion exhaustion.

The maintained [`regex` package documents per-operation timeouts](https://github.com/mrabarnett/mrab-regex#timeout).
The Dataiku project-library SDK returns whole files rather than ranged streams,
so the accepted-file limit applies after the SDK response; an in-flight request
remains governed by the Dataiku client's HTTP timeout. The PR states that
constraint plainly instead of claiming a range the SDK does not expose.

## Evidence and provenance

- PR #4 introduced the fixed surface in
  [`c753f9b`](https://github.com/dataiku/dku-headless/commit/c753f9b75e32bf59521b8e46c5df23e2c7d9f04d).
- Its generic inspector came from
  [`9edf6fd`](https://github.com/dataiku/dku-headless/commit/9edf6fd2) and
  [`610787a`](https://github.com/dataiku/dku-headless/commit/610787a7).
- The [scope comment](https://github.com/dataiku/dku-headless/pull/4#issuecomment-5028205473)
  asked for small changes whose trade-offs can be reviewed on their own. This is
  the isolated pruning decision.
- [Anthropic's tool-design guidance](https://www.anthropic.com/engineering/writing-tools-for-agents)
  argues for parameterized consolidation instead of overlapping narrow tools.
- GitHub's MCP server made a
  [comparable default-tool reduction](https://github.com/github/github-mcp-server/discussions/1182).
- The [MCP tool specification](https://modelcontextprotocol.io/specification/2025-06-18/server/tools)
  defines the advertised names and schemas as the discoverable server contract.
- Dataiku's [project API](https://developer.dataiku.com/latest/api-reference/python/projects.html)
  and [dataset API](https://developer.dataiku.com/latest/api-reference/python/datasets.html)
  support the underlying read paths. The locked client source was checked at
  [`137e7669`](https://github.com/dataiku/dataiku-api-client/commit/137e7669a5423166040319dff11dc5f94f6ccbbe).

This slice is not a `dku-cli` port. It comes from the named PR #4 commits, the
review comments, the current SDK, and the tool-design sources above. The later
skill PRs document their separate `dku-cli` provenance.

## Risk controls

- The exact catalog is an allow-list test, not a count-only assertion.
- Every removed generic path has dispatch/equivalence tests.
- Dataset column selection has a direct regression test.
- Library source filtering, depth, UTF-8 clipping, malformed text, oversized
  files, file counts, regex timeout, syntax failure, and final byte ceilings have
  focused tests.
- Legacy transport gates remain unchanged and are tested in subprocesses.

## Verification

- Python 3.10: **245 passed**
- Python 3.13: **245 passed**
- `uv run ruff check .`
- `uv run cz check --rev-range agent/pr4-09-fixed-surface..HEAD`
- `git diff --check`

## Stack

Draft, based on #33. Review #33 first; GitHub will then show only this pruning
and hardening slice. The next draft replaces the temporary per-object skills
with the single routed skill.


crmapj:
Closed in the rework: the reader prune now lands atomically with the inspector in #31. The library hardening half was dropped per the design review; the surface prune to 53 tools is #39.

PR #44, opened 2026-07-21
Refactor/mono skill

pmasiphelps:
Changing skill structure to be mono-skill plus references. Super lightweight refactor.

- Look at commit 1 for rename/file reorg only
- Look at commit 2 for actual skill file changes only

PR #45, opened 2026-07-22
feat(projects): restore set_project_variables

pportal2:
Cobuild cannot write project variables, forcing migration agents to hardcode values such as reference dates in recipes. This restores `set_project_variables` from agent-dev-kit with its original wholesale replacement contract, adds five unit tests, and exposes `get_project_variables` in FULL Cobuild mode for read-modify-write workflows.

This PR addresses this Trello card: https://trello.com/c/pNqhhn2f/150-set-project-variables-tool


pportal2:
Waiting for #33 and rebase before merging

nmartorell:
Thanks @pportal2 ! I reorganized the fake client a bit, and added explicit tests to it, but feel free to revert or comment if you don't like something. 

My only other comment here is that the README and project skills document needs to be updated to reflect that we are adding "set_project_variales" as a mutating call due to Cobuild not supporting it, but we should do this after merge of #33 (as this PR changes the README file section that we need to update).

pmasiphelps:
I updated the project skill doc in #44 to reflect the new set_project_variables direct tool. I think we only need it here, not in the readme.

nmartorell (approved):
Code changes look good. 
Make one small change to add the new set_project_variables method to the mcp tool list test.

pmasiphelps (approved):
looks great!

PR #57, opened 2026-07-23
refactor(flow)!: replace get_flow_items_in_traversal_order with get_flow_graph

crmapj:
Replaces the item-by-item get_flow_items_in_traversal_order tool with a single get_flow_graph tool that returns the whole flow in one call.

get_flow_graph returns nodes as [ref, short_type] pairs, edges as [from_ref, to_ref] pairs, a source count, and an ASCII build tree. The tree reads top-down as build order: each source dataset is a root, nesting is downstream dependency, sibling branches are parallel paths, and a leaf marked (^) is a re-convergence point drawn once instead of re-expanded. You can scope to a single flow zone with the zone argument.

Flows really do get large, so the response is bounded. Nodes cap at 2,000 and edges at 10,000, callers can raise the defaults within those hard ceilings or narrow by zone. When the returned graph is partial, or the flow exceeds the tree threshold, the tree is omitted and the caller relies on nodes and edges, with a warning explaining why.

This ported from source commit d16d256. That commit also carried get_project_overview and a redaction refactor, both deliberately left out here as a separate concern so this stays a focused surface change.

Adaptations relative to the source: ported onto current main rather than cherry-picked, since the source branch sits on stale bases. The flow-graph tests were lifted out of the source test_context_tools.py into a new self-contained tests/test_flow_graph.py. The pinned tool surface contract test and the skill reference docs were updated to name get_flow_graph.

Review hardening, round 1 (second commit), fixing three defects inherited from the source commit:

1. Zone scoping now resolves membership server-side through DSSFlowZone.get_graph() instead of reconstructing it from raw zone item lists. The old filter returned an empty graph for the implicit default zone, whose items list is empty by design, and ignored shared and foreign-project items.
2. The ASCII tree now has its own 200,000-character rendering budget. The node and edge ceilings alone still allowed multi-megabyte trees on dense graphs, since the tree emits one line per edge traversal with a depth-sized prefix. Past the budget the tree is omitted with an explicit warning.
3. Cycles are flagged distinctly from re-convergence. Back-edges to a node still open on the current traversal path get their own marker and a warning that the flow data is malformed, instead of being mislabeled as an ordinary join.

Review hardening, round 2 (third commit), closing the residual gaps:

1. The serialized response itself is now capped at 1,000,000 characters, measured on the final JSON string so escaping cannot blow past it. Past the cap the response degrades in stages, tree first, then the edges list, then the nodes list, each keeping a stable prefix with original counts and explicit truncation warnings. The reviewer's 299-node, 10,000-edge shape with maximal refs went from 10.5 MB to about 800 KB. The tree also switched to ASCII-only characters, since box-drawing glyphs double or triple under JSON escaping.
2. Zone resolution is two-pass: an exact id match wins across all zones before any name match, so a zone name can no longer shadow another zone's id, and a name shared by several zones is rejected with the candidates listed.
3. Cycle detection is its own iterative DFS over the returned graph, decoupled from tree rendering, which can abort on its character budget. The cycle warning is now complete regardless of traversal order and uses the same text whether or not the tree rendered.

Review hardening, round 3 (fourth commit), making the guarantees absolute:

1. The response ceiling now holds for any input, not just large graphs. After the list-degradation stages, remaining oversized echoed strings (the unbounded project_key) are clipped with truncation metadata, the tree title is clipped at render time, and a wholesale envelope backstop, unreachable by construction but kept for provability, truncates the serialized payload as a last resort. A 1,000,000-character project_key on an empty graph now returns 775 characters instead of slightly over the ceiling.
2. Count fields are now internally consistent under clipping. node_count, edge_count and source_count all describe the full zone-scoped graph computed before any clipping, and returned_node_count and returned_edge_count describe exactly what the response carries. Previously node clipping could leave source_count at an intermediate subset size true of neither graph. The docstring states which graph each count describes.

This removes a public MCP tool, so it is a breaking surface change.


pmasiphelps (approved):
@crmapj this looks good! 

I made a few changes in my commit. Please take a look!

1. Kept the useful graph data in the response (node_types + edges), but dropped the ASCII tree since it was redundant and made responses larger. On projects in the agent dev kit instance, this cut token usage by about 40% compared with the earlier get_flow_graph shape.
2. Removed per-string clipping of individual dataset/recipe/model/etc. refs. Our other list_* / get_* tools do not do this, so it did not feel necessary here either.
3. Kept whole-response size protection, with explicit truncation metadata when edges or node_types have to be clipped.
4. Fixed cycle warnings so they reflect the full scoped flow, even when the returned graph is clipped.
5. Cut down the tool docstrings and inline comments. Moved usage guidance into the skills (projects.md / recipes.md) instead of keeping it in Python docstrings.
6. Updated truncation guidance to recommend targeted list_* / get_* follow-up, not zone scoping, since large flows are not necessarily organized into useful zones.

crmapj:
Thanks @pmasiphelps — agreed on all of it, and the tree removal is worth more than you claimed. I benchmarked both commits against a real instance (2,155 project flows) rather than synthetic graphs, and pushed three follow-ups in f42ef3a.

## Your change, measured

Running `4e48675` and `0ef44c9` head-to-head over the twelve largest real flows:

| project | nodes | before | after | cut |
|---|---|---|---|---|
| EPIROCSUPPLYCHAINFORECASTING | 289 | 61,966 | 37,828 | 39% |
| SOL_DEMAND_FORECAST_1 | 153 | 40,566 | 18,396 | 55% |
| DSSUSAGEINSIGHTS | 148 | 33,583 | 16,699 | 50% |
| SOL_CUST_SEG_BANKING | 126 | 29,850 | 13,593 | 54% |
| BB_DC_MODEL | 120 | 24,563 | 9,020 | 63% |
| SOL_MBA | 99 | 25,135 | 12,314 | 51% |

Mean ~51%, not ~40%. The tree was 38–63% of every payload, and since the tree threshold was 300 nodes while the biggest flow on the instance is 289, it rendered on **every** flow — so that saving landed on every call. You were also right about the docstrings on stronger grounds than stated: median tool docstring in this repo is 9 words, the longest is `build_datasets` at 83, and the version you replaced was ~190.

## Three follow-ups

**1. `nodes_by_type` instead of a ref-keyed map.** A ref is only unique within one object type — DSS datasets and recipes are separate namespaces, and `dataikuapi` itself identifies a graph node by `(type, ref)`. So the ref-keyed dict silently dropped nodes:

```json
"node_count": 3, "returned_node_count": 3,
"node_types": { "orders": "recipe", "out": "dataset" }   ← 2 entries
```

That breaks the count invariant 4e48675 was specifically written to establish. Grouping keeps both nodes, and because it stops repeating the type once per node it's also **8% smaller** than the map — so this is a token win first and a bug fix for free.

**2. `sources` is now named, not just counted.** Sources can't be rederived from the returned edges once those clip: a node whose only in-edge was dropped looks like a root. On a 900-node flow with edges clipped, deriving from `edges` gives 500 sources against a true `source_count` of 300 — two irreconcilable answers in one response. `sources` is full-graph truth like the counts and degrades last under the budget. Costs +0.9%; the grouping above pays for it four times over.

**3. Edge ceiling lifted 2,000 → 5,000.** `_MAX_GRAPH_EDGES` and the `max_edges` default were both 2,000, so the knob could only be turned *down* — and this warning was unfollowable:

```
"edges: returning 1600 of 2000; increase max_edges within the documented ceiling..."
→ max_edges=4000
→ ValueError: 'max_edges' must be <= 2000
```

Real flows run up to ~1.7 edges per node (measured), so a flow at the 2,000-node ceiling needs ~3,400 edges. The ceilings shouldn't be equal.

I also folded the two duplicated halving loops in `_fit_response_to_budget` into one helper, since adding `sources` as a third degradation stage would otherwise have meant a third copy. Net effect: **−7.3%** versus your shape, despite adding a field.

## Caveats on my own numbers

The instance I tested is a sandbox: 2,155 flows, biggest 289 nodes, zero ref collisions, nothing ever clipped. So items 1 and 3 fix bugs that are **latent there, not live** — I'm fixing them because customer production flows do reach 1,000+ nodes, not because I observed failures. Item 2's clipping scenario likewise needs a flow ~4× bigger than anything on this instance.

Verified: full suite green (107 tests), ruff clean, and the invariants (`returned_node_count == len(nodes_by_type)`, `source_count == len(sources)`, `sources == graph roots`) hold across a 400-flow live sweep.

One thing I left alone deliberately: the count-ceiling warnings still say "or scope by zone" while the budget warnings say "use the relevant list/get tools". Your reasoning for dropping zone-scoping advice applies to both, but the wording is your editorial call — say the word and I'll harmonise.

pmasiphelps (approved):
Changes look great. I just tightened up some code comments. I think claude is over explaining its changes to code, which shouldnt end up in the final codebase.

PR #59, opened 2026-07-23
feat(execution): add build_datasets, run_recipe, run_scenario

crmapj:
Add three direct-execution MCP tools for existing assets: build_datasets, run_recipe, and run_scenario. These are the fixed deterministic executions of existing Flow assets. Everything that creates or modifies a project asset still routes through Cobuild.

The implementation is ported from commit 70c7389 (draft PR #29), rebuilt onto current main since that branch was based on a stale stacked chain.

What the tools do.

- All three default to fire-and-return, with an optional bounded inline wait.
- build_datasets builds every requested output in one job, so there is one job_id and one shared timeout.
- run_recipe triggers a recipe by building its first output, so auto-update-schema attaches to the job builder.
- run_scenario returns the trigger fire id until DSS materializes the real scenario run, so the caller never receives a misleading run id.
- Every post-start outcome keeps the job or run identity in the response, so a timeout never invites a blind duplicate run.

The wait-doc wording follows PR #40: timeout_seconds is a soft deadline checked between SDK polls, not inside them, and a single hung DSS HTTP call can exceed it.

The three tools are pinned into the fixed tool-surface contract test. The jobs and scenarios reference guides, the SKILL.md routing table, the README tool-surface summary, and the Cobuild write-routing convention are updated to match.

Review hardening. Round 1 found run_scenario could drop the trigger identity on timeout-before-run and on cancellation (fixed in 0d89f2d). Round 2 found post-start exceptions could lose the identities under FastMCP error masking and that run history could not correlate a trigger id (fixed in 2711a77: structured poll-failed payloads instead of raising, run_id kept when observed, trigger_fire_id exposed in history). Round 3 hardened the handlers in 79a0b3d: safe error-text rendering, defensive scenario run-id capture, shape-checked history trigger data. Round 4 in f20a48d gave build_datasets and run_recipe a guarded job-id capture. Round 5 in 001bb82 fixed the deeper root cause: _wait_for_job_result itself read job.id, so a handle whose .id raises while get_status returned a terminal DONE payload collapsed a completed build into a poll failure; the helper now takes the caller's captured job_id and never reads the handle, and run_scenario likewise captures trigger_fire_id once (guarded) right after run() and reads the fire handle nowhere else. Round 6 in 78d9787 made the always-carries-the-field contract literally true: all five post-start payloads now emit trigger_fire_id as a key (scalar id when known, null when a partial handle would not yield it) rather than omitting it, so a client never distinguishes an omitted key from a null.

Which tests genuinely fail pre-fix. The three hostile-handle regression tests fail against the pre-fix source and pass after: build_datasets and run_recipe with a handle whose get_status returns a terminal DONE payload while .id raises (verified against 001bb82's parent), and run_scenario with a trigger-fire handle whose .run_id raises. The boolean, list, and absent-runId history-shape tests are coverage of already-correct behavior, not regression proofs: the isinstance scalar filter has been in place since round 4, so those three pass against earlier commits too. The test file carries a comment stating this so the record is accurate.

trigger_fire_id contract (on record). get_scenario_run_history follows one fixed rule: trigger_fire_id is the fire record's runId only when that runId is a string or integer, booleans excluded (a boolean is a JSON scalar but never an id), otherwise null; trigger_type is read only when the inner trigger is a mapping, otherwise null. A valid string or integer fire id is kept even when its sibling trigger definition is a bare string. A non-scalar, boolean, or absent runId is always null and never a fabricated structure. run_scenario post-start payloads always carry the trigger_fire_id key (scalar when known, else null). The string, integer, boolean, list, dict, and absent shapes are each covered by a test.

Live verification. The branch was test-driven against a real DSS 14.x sandbox (project AAA_144D45): build_datasets no-wait plus polling, build_datasets and run_recipe bounded waits, and every run_scenario path work live, including trigger-id correlation through get_scenario_run_history. The drive exposed one real bug the mocked tests missed, fixed in 934b739: live DSS keeps an activity's output refs at def.targets as datasetName entries (no top-level targets, empty statusOutputs even when DONE), so the activity summarizer in job_summaries.py produced empty outputs and per_dataset came back all null on a completed DONE build. This shape bug is latent on main already, get_job_status full=true exposes the same empty outputs; the new per_dataset field surfaced it. The summarizer now reads def.targets with a fallback to the old shape, the test fakes mirror the captured live payload, and a regression test pins per_dataset DONE on that payload.

Validation. Full pytest suite passes (81 tests) and ruff is clean.


pmasiphelps (approved):
@crmapj looks great! i made a few changes:
- Trimmed the new tool docstrings so they match the rest of the repo better. Made sure this guidance was in the reference docs.
- Tightened a few comments and helper docstrings that were explaining the implementation change in too much detail.
- Added a note in recipes.md to clarify Cobuild-driven execution during broader project work and one-off execution through jobs.md (does that look right to you?)
- Moved the run id/trigger id scenario explanations to a dedicated section in the skill doc. It felt out of place in the Workflow section.

PR #60, opened 2026-07-23
feat(cobuild): consent, retained turns and pollable status

crmapj:
Improves Cobuild conversation management by combining the reworked consent layer (source commit 8ca6a15) and process-local retained turns (source commit ea02ea1), plus the atomic instance binding and bounded retention from PR #40, ported onto current main.

## Consent

`send_cobuild_message` now defaults `allow_edit_project` to false. Editing is opt-in per message and the grant does not carry to later messages. Deletion proposals carry their `confirmation_id` alongside `objects_to_delete` and `deletion_impacts`, and `answer_cobuild_confirmation` requires that exact id before it will approve or cancel, so an agent cannot blind-approve a proposal it never read. Conversation listings expose `has_pending_confirmation` as a boolean and never the raw token.

## Retained turns and pollable status

Each turn runs in its own named daemon thread and survives a client timeout. On timeout the caller gets `status=timeout` with a `turn_id` instead of losing the turn, and polls the new `get_cobuild_turn_status` tool to the real outcome. The rule is poll, never re-send a timed-out mutation. One turn per conversation runs at a time: a second send returns `status=busy` with the existing `turn_id`. A global cap of 3 concurrent turns returns a saturated error when full.

## Binding and retention (from PR #40)

Instance name and DSS client are captured in one atomic snapshot at tool entry, before work is handed to the daemon thread, so a concurrent `switch_instance` cannot pair one instance's name with another's client. Settled turns are swept by TTL and count, and idle unconfirmed conversations are capped. A turn in flight or an armed deletion confirmation is never dropped.

## Review hardening (round 1)

An adversarial review found four defects, all fixed.

Orphaned turns are recoverable through the public surface. A send whose MCP request is cancelled used to leave the accepted mutation undiscoverable, so a retry after settlement silently ran a second SDK mutation. Each conversation now retains its latest `turn_id`, `list_cobuild_conversations` exposes `last_turn_id` plus `last_turn_status`, and `get_cobuild_turn_status` polls the conversation's latest turn when `turn_id` is omitted. The documented recovery path after an interrupted send is list-then-poll, never a blind resend, and a regression test recovers a cancelled turn using public tool calls only.

The settled-turn sweep can no longer evict an armed deletion proposal. While a proposal is unanswered, its settled turn is pinned outside both the TTL and the 64-turn cap, so the exact `confirmation_id` with its objects and impacts stays retrievable and answerable. It becomes sweepable once answered or cancelled. The cap now counts unpinned turns only.

The terminal-payload bound from the source design is restored and applied in the worker thread before retention, so an oversized result never stalls the event loop or accumulates in the registry. An oversized deletion proposal is disarmed, never truncated. An ordinary oversized response is clipped with explicit truncation metadata (`truncated`, `original_bytes`).

The 128-conversation cap is enforced on the public path. `start_cobuild_conversation` evicts the oldest idle entry to make room and refuses with a clear error when nothing is evictable. Eviction protects the entry being admitted or about to run a turn, so sending on the oldest conversation can never evict it mid-use.

## Review hardening (round 2)

A second review pass found six further issues, all fixed.

A conversation whose latest turn was never delivered to any client is no longer cap-evictable, so a cancelled, settled mutation stays discoverable even when creations churn the registry at the cap. Delivery is tracked whenever a payload carrying the `turn_id` reaches the client (terminal, timeout, or busy result, or a status poll). If that leaves nothing evictable at the cap, creation refuses with a clear error instead of dropping the unobserved conversation.

The window between entry resolution and turn activation is closed. Resolution now pins the entry inside the same lock acquisition, and the pin is released once the turn is active, so a concurrent creation can no longer evict the conversation a send is about to use. A regression test interleaves a creation into exactly that window.

Disarming an oversized deletion proposal now actually answers CANCEL to DSS in the worker, using the same SDK call the answer tool uses, and reports `cancelled_confirmation=true` only when that call succeeded. If the DSS cancel fails, the payload honestly reports `confirmation_disarmed_locally=true`, `cancel_failed=true`, and `cancel_error`, notes the conversation may still await confirmation server-side, and keeps the local rejection of the id.

Admission now runs before the SDK creates the remote conversation, reserving the slot under the lock, so a refused creation never leaks a server-side DSS conversation. A test asserts a refusal makes zero SDK creation calls.

The 512 KB ceiling is now absolute. Replacement payloads are re-measured after construction and their embedded server-supplied strings (conversation id, error text, message) are hard-clipped until the serialized result is provably under the ceiling, closing the pathological-conversation-id hole. The skill guide documents the exact emitted fields.

Recovery docs and the `get_cobuild_turn_status` docstring now direct callers to poll the exact `last_turn_id` obtained from a listing. The omitted-id form is documented as a convenience that polls the current latest turn, whose returned `turn_id` the caller must compare, since a newer turn may have started after the listing.

## Review hardening (round 3)

A third review pass found six edge-case issues, all fixed.

The settled-turn sweep now also exempts a conversation's still-unobserved latest turn, not just armed deletion proposals. A cancelled, settled mutation is therefore never swept out of the turn registry while its conversation is still pinned, so the listing keeps a `last_turn_id` that polls to a real outcome. This is bounded: at most one unobserved latest turn per conversation, and conversations are capped, so it cannot grow without bound. Once the turn is observed through any public poll it sweeps normally.

The `missing_confirmation_id` terminal branch now routes through the same size bound as every other terminal payload, so a pathological `conversation_id` on that path can no longer return or retain an oversized result.

The SDK conversation object's own retained message history is now bounded. The mechanism is a best-effort prune after each turn: any retained history entry whose serialized size exceeds the terminal bound has its bulky content replaced with a truncation marker, in place, in the worker thread. The prune touches an SDK-private message list, so every step is guarded and tolerates a missing or renamed attribute or an unexpected entry shape, never crashing the turn. Repeated oversized turns on one conversation therefore cannot grow retained memory without bound.

Disarming an oversized deletion proposal no longer erases a successor. After the CANCEL SDK call, the conversation's actual pending state is re-read rather than forced to None. If the cancel raised a new, smaller confirmation request, the response surfaces `successor_confirmation_pending=true` with its `pending_confirmation_id` instead of claiming a clean cancellation, and the successor stays armed. Pending is forced to None only when the cancel failed, which keeps the oversized id locally rejected.

Admission pins are now token-based and taken synchronously immediately after resolution, with release in a finally on both the send and answer paths. Cancelling a request during or after the off-thread resolution can no longer leave a pin stranded and a conversation stuck non-evictable, because the release removes exactly that request's token and never another's.

Conversation creation now defers eviction until the remote conversation exists. Reservation only checks feasibility before the SDK call, and the evict-and-insert commit runs after a successful create, so a definitive create failure at the cap no longer destroys a pre-existing local conversation while also creating nothing in DSS.

## Review hardening (round 4)

A fourth review pass found three further issues, all fixed.

A CANCEL of an oversized proposal that raises a successor confirmation now captures that successor's full response and returns it as a first-class `status=needs_confirmation` result, with its own `confirmation_id`, `objects_to_delete`, and `deletion_impacts`, plus `superseded_oversized_proposal=true`. Because the status is `needs_confirmation` and the id matches the conversation's pending id, `_turn_is_pinned` treats it as armed, so it is inspectable, protected from the count and TTL sweep, and answerable by its exact id, instead of the previous id-only error payload that the sweep could drop while the conversation stayed pending. It is bounded like any proposal, with a recursion guard for pathological chains.

The retained-history prune now measures the real SDK object. The dataiku-api-client history entries are `CobuildAssistantResponse` objects whose large payload lives under `_raw` while `str()` is only a short summary, so the previous str-based sizer measured a 600 KB proposal at under a hundred bytes and left it intact. **Mechanism:** sizing now reads `_raw` (and defensively `raw` and the instance `__dict__`) and serializes that to get the true size; when an entry is over the bound, the prune clears its bulky `_raw` payload in place on the object, preserving object identity, so repeated oversized proposals keep retained memory bounded. Every step is guarded so an unexpected shape never crashes the turn.

Conversation admission now re-validates the cap as a hard invariant at commit, not only at reservation. If the entry that was going to be evicted has become active between reserve and commit, the commit evicts an actually-evictable entry, or, if none is evictable, refuses the creation instead of inserting over the cap. A public-tools regression drives exactly that interleaving at cap 1 and asserts the registry never exceeds the cap.

## Scope

Everything is process-local. There is no durable store on disk, no lock files, no cross-process coordination, and no credential fingerprints. A restart means a new conversation, and that is the accepted v1 design.

This ports the behavior of the source PRs rather than their code, which was stacked on an unmerged runtime-safety PR and a pruned tool catalog. `get_dss_client` gained an optional `instance` argument so the captured snapshot builds its own client, and `get_cobuild_turn_status` was added to the pinned tool surface contract test.


nmartorell:
Closing as it this PR is superseded by #78 .

PR #64, opened 2026-07-24
fix(managed-folders): support managed folder creation

pportal2:
Fixes managed-folder skill that previously routed folder creation through Cobuild, which cannot create managed folders. Adds create_managed_folder so agents can create one directly using a caller-selected connection.

https://trello.com/c/FEXdDSjM/163-create-managed-folder-tool


pmasiphelps (approved):
Looks good! i added a couple doc tweaks

PR #65, opened 2026-07-24
feat(datasets): add complete local export

pportal2:
Adds export_dataset, a Headless tool that writes every row from a DSS dataset to a UTF-8 CSV on the machine running the stdio server. It supports ordered column selection and explicit overwrite, writes through a temporary file to avoid partial outputs, and returns the output path, row and column counts, schema types, file size, and SHA-256. The dataset skill now directs agents to use it when a complete local export is needed.

https://trello.com/c/YhDhxeyc/164-export-dataset-tool


pmasiphelps:
@pportal2 what is the use case for the export tool? id rather we increase the limit on the get_dataset_sample tool, and make a human manually click an export dataset button

pportal2:
@pmasiphelps: open to discuss this and figure out the right approach.

The use case comes from migration work: headless needs a reliable way to compare a Dataiku dataset with a local "ground truth" file.

Increasing the limit cap of get_dataset_sample won't solve this because streaming dataset content into the agent context window doesn't allow for programmatic comparison. Writing to a local file lets the agent use Python locally to compare the two.

Other alternatives:
- give get_dataset_sample an optional output_path parameter that saves the tool output to a CSV file when specified
- do it directly in the Dataiku Flow by asking Cobuild to run the comparisons

pmasiphelps (approved):
@pportal2 looks great. i just made a few changes as discussed on our call today.

1. Added a limit param to export_dataset and enforced a hard cap of 1,000,000 rows.
2. Changed the export result from truncated to has_more_rows, which is a better fit for an intentionally bounded export.
3. Added docs guidance in datasets.md that large exports can take a while and that limit should be used to bound them.
4. Tweaked the datasets.md description so the export tool is only mentioned at the end of the description, not in the opening sentence (so as to not steer an agent too much?).
5. Validated the updated behavior against the live dev-dss instance with a small real export.

PR #69, opened 2026-07-28
review: feat/dku-onboarding-and-plugins branch

nmartorell:
Hi @yjouini!

It's late so the description is short :) but I wanted to share in case you had some time to look at these suggestions before standup tomorrow. No worries if not though, we can review during (or before/after if you have time).

Overall, I think your approach is great. The suggestions in this branch are mainly due to the fact that the code in config.py wasn't really originally planned for both reading and writing of config files, so the result was a bit awkward. Specifically, there were a few code paths that lead to possible staleness of the in-memory cache of instances. So I thought it would be easiest to just get rid if it.

The suggestions in this branch are mainly:
1. Get rid of the in-memory cache of instances (only the current instance is in-memeory); reading small config files from disk is cheap, and we will only do this very seldomly anyway (i.e. when swithcing or listing instances); the upside is we have no staleness issues.
2. Also in config.py: now we have just one path to discover config.json; the idea is that (a) explicit always takes precendence, (b) if you're in dev mode you default to read/write of cwd, and (c) if nothing else we default to config.json in home dir; in the rpevious implementation, if working in dev mode, adding an instance would result in the cwd config.json being compied over to the home dir; not sure if that was intended, we can discuss if so.
3. More "agent-centric" error messages: removed the references to settings env vars and manually editing config.json, as these are not things that the agent will be able to do (without hackily restarting the mcp server); the idea is that these are both "advanced" or CI features that we can expect users to read the docs for.

Anyway, this is really all just backend-dy stuff, overall very much agree with the approach, thanks for coming up with it!

p.s. finished a bit late so need to still run some manual integration tests on this branch, just fyi in case you run this and it blows up ;)


yjouini (approved):
Thanks Ned ! I went through the changes and agree with most of them. 

I pushed a small follow-up fixing active-profile deletion: deleting the current persisted profile now switches to the saved default, or clears the active instance when none remain. I also replaced the path-like source marker with explicit config/environment source labels and added regression tests.

Tested it, LGTM !

nmartorell:
@yjouini thanks for reviewing the review :)

I agree with the proposed changes as well - I went back and forth with what to do about the side-effects of deleting the active instance, I think your solution works nicely.

PR #72, opened 2026-07-28
refactor: remove PyPI

hanneshapke:
This PR removes the PyPI release flows and updates the documentation. 

Call out: Without the PyPI package, the invocation changes. I am falling back to 

```
      "command": "bash",
      "args": ["./bin/run_mcp.sh"],
```


opencode.json
```diff
@@ -0,0 +1,10 @@
+{
```
hanneshapke: @yjouini @pmasiphelps Should this file be removed?

pmasiphelps: why remove it?

.cursor/mcp.json
pmasiphelps, line 1: why did you include this file? i believe the cursor install will happen via cursor CLI per the README: `cursor agent plugin marketplace add github.com/dataiku/dataiku-headless`

pmasiphelps:
@hanneshapke i made two minor changes:

1. README.md - changed `.mcp.json` instructions to use "command": ["bash", "./bin/run_mcp.sh"] instead of "command": ["uvx", "dataiku-headless", "serve"]
2. RELEASE.md - changed the explicit list of plugin manifests to just refer to the `pyproject.toml` list of plugin manifests

opencode.json
pmasiphelps: i am confused with this. i think you're right to remove it (as well as the .cursor/mcp.json file introduced in this PR), but i want @yjouini 's take on this :)

yjouini: Yes both this and .cursor/mcp.json should be removed. They are only applied when opening a cloned repository directly and not intended for plugin distribution

yjouini (approved):
LGTM after removing opencode and cursor files !

PR #73, opened 2026-07-28
fix(cobuild): support ask-question requests

pportal2:
dataiku-api-client 14.7.2 adds support for Cobuild question turns. Previously, the MCP allowed client versions from 14.7.0 and only serialized response.message. That field is null on question turns because the prompt is stored in title, leaving agents without the question text or a way to respond.

This PR requires dataiku-api-client>=14.7.2, includes question details in serialized responses, and adds answer_cobuild_question for answering or declining through the correct endpoint. It also updates the Cobuild skill reference and tool-surface contract test.

Also replaced the generic object annotation with DSSCobuildConversation.

https://trello.com/c/5TPWqsBu/166-add-ability-to-answer-cobuild-questions


nmartorell (approved):
Thanks for staying on top of the Cobuild API @pportal2 😄 this is very timely as I'm currently reviewing the whole Cobuild module.

PR #74, opened 2026-07-28
feat: mcp with loading specific Python version and dependency

hanneshapke:
Using `manzt/uv@0.8.13` to load uv via npx, uv is then setting the Python env with version and deps.


nmartorell (approved):
Tested on codex and claude code, works perfectly. 

One caveat is that due not yet having a tag for any releases, I had to test as a "local filesystem" marketplace instead of allowing codex/claude to pull the repo directly from gh. This should be something we should test once we merge to main and create our first release.

How "interesting" that claude code doesn't accept standard commitizen tag from ☹️ nice catch!

bin/launcher.sh
```diff
@@ -0,0 +1,154 @@
...
+# The first tier that works becomes the server process. If none do, we exit
+# non-zero telling the user to install uv.
+#
+# stdout is the MCP JSON-RPC stream; everything logged here goes to stderr.
+
+set -eu
+
+HERE=$(CDPATH='' cd -- "$(dirname -- "$0")" && pwd)
+SERVER="$HERE/run_mcp.py"
+NPM_UV_PACKAGE="@dataiku/uv@0.12.0"
```
hanneshapke, line 25: @nmartorell :)

pmasiphelps (approved):
Tested both claude code and codex plugins on my mac with uv and npx installed using the local file path rather than github repo link - works great!

yjouini (approved):
Hey @hanneshapke Thank you the new solution works for me! Minor concern I could not test: Python fallback still might be broken on native Windows? I think the manifests require sh, and the venv path is hardcoded to venv/bin/python instead of Windows’ venv/Scripts/python.exe. Should we handle this in another PR ?

PR #76, opened 2026-07-30
fix(onboarding): normalize Dataiku instance URLs

crmapj:
## Summary

- accept any well-formed HTTP(S) page URL from a DSS instance
- normalize paths, queries, fragments, and trailing slashes to `scheme://host[:port]` before saving
- reject malformed URLs, credentials, invalid ports, whitespace, control characters, and backslashes
- explain in the form that any Dataiku page URL can be pasted
- cover common UI routes and adversarial URL shapes

## Why

Users commonly copy the URL of the Dataiku page they are viewing, such as `/home`, `/flow`, or `/projects/...`. The API client needs only the DSS base address. Parsing one URL and discarding its page-specific parts is simpler and more durable than maintaining an allowlist of UI routes.

## Review feedback addressed

A pasted `/flow`, `/projects/...`, or other Dataiku page URL is now accepted and reduced to its origin instead of being rejected. The custom browser regex and server route allowlist have been removed in favor of the standard URL parser.

## Validation

- `uv run pytest -q` — 134 passed
- `uv run pre-commit run --all-files` — passed
- Python syntax check for `dataiku_mcp/setup_server.py` — passed
- `uv run cz check --rev-range origin/main..HEAD` — passed


dataiku_mcp/setup_server.py
```diff
@@ -56,6 +58,12 @@ def _validate_form(form: dict[str, list[str]]) -> dict:
         raise ValueError(
             "Instance URL cannot contain credentials, a query, or a fragment."
         )
+    if parsed_url.path not in ALLOWED_INSTANCE_URL_PATHS:
+        raise ValueError(
+            "Instance URL must be the instance root or its /home page, "
+            "without any other path."
+        )
```
yjouini: why ? if i paste instance_url/flow or any other instance path starting with the right instance url it should be fine, juste need to parse it

yjouini (changes requested):

yjouini:
Should it fix the url or tell the user it's not the right one ?

yjouini (approved):

PR #78, opened 2026-07-30
feat: manage Cobuild conversation turns

nmartorell:
## What This PR Does

This PR provides a deliberately small, process-local conversation manager for Cobuild.

- Retains one current turn per conversation. A send, confirmation answer, or question answer starts background SDK work; the tool waits up to 240 seconds, then returns the current \`turn_id\` and \`queued\`/\`in_progress\` if work continues.
- Serializes operations per conversation. A new operation is rejected while the current one runs, or until its terminal result has been retrieved through \`get_cobuild_turn_status\`.
- Uses a dedicated Cobuild \`ThreadPoolExecutor\`, configured by \`DKU_MCP_MAX_COBUILD_WORKERS\` (default \`4\`), so long-running Cobuild work does not consume ordinary MCP blocking workers. Excess turns queue in that executor.
- Makes confirmation and question responses first-class retained-turn states. Both are answered only against the exact current \`turn_id\`; question results expose their title and answer constraints. The wrapper does not expose or directly depend on the SDK's private confirmation ID.
- Defaults \`allow_edit_project\` to \`false\`; edits remain opt-in per message.
- Provides recovery-oriented status and error messages, plus agent skill/tool documentation that directs callers to retrieve and inspect a retained result before answering.
- Updates \`dataiku-api-client\` to \`14.7.2\`, which provides the public question-response SDK API used here.

## Comparison With PR #60

PR #60 (\`feat/cobuild-conversation-management\`) tackles the same reliability problem with a substantially more comprehensive retained-turn registry. It provides several protections this PR intentionally does not:

| Area | PR #60 | This PR | Tradeoff |
| --- | --- | --- | --- |
| Turn history and retention | Retains independently addressable turns, with TTL/count sweeping and recovery of older turn IDs. | Retains only the current turn for each conversation. Once a new turn starts, the previous turn is no longer queryable. | PR #60 offers stronger historical recovery and bounded retention; this PR has a much smaller state model, but conversation entries can grow without a registry cap. |
| Memory and payload bounds | Bounds terminal payloads and SDK history; truncates ordinary large results and disarms oversized deletion proposals rather than exposing an uninspectable approval. | Does not impose payload or SDK-history limits. | PR #60 is safer for long-lived, high-volume, or pathological responses; this PR avoids complex truncation, SDK-private history mutation, and automatic cancellation behavior. |
| Capacity policy | Uses a global three-turn semaphore and returns a structured saturation response when full. | Uses a dedicated worker pool and queues additional Cobuild turns. | PR #60 puts a hard bound on active work and memory; this PR avoids asking an agent to retry after capacity rejection, at the cost of potentially long and unbounded queue delay. |
| Timeout/status model | Per-call configurable timeout (\`1200s\` default, \`1800s\` maximum); status is a snapshot poll and can retrieve retained historical turns. | Fixed \`240s\` bounded wait for every send, answer, and status call; status only accepts the current exact turn. | PR #60 is more configurable and supports historical polling; this PR reduces agent polling and API surface, but offers less caller control and no prior-turn lookup. |
| Confirmation identity | Returns and requires the SDK's exact private confirmation ID; fails closed if it is absent or mismatched. | Uses the exact current wrapper \`turn_id\`; validates that its retained result is a confirmation before calling the SDK. | PR #60 binds directly to DSS's proposal token; this PR avoids exposing a private SDK implementation detail while still rejecting stale or duplicate answers through current-turn serialization. |
| Instance/concurrency hardening | Captures the instance/client binding atomically, uses registry locks and admission pins, and protects against races during eviction or instance switching. | Relies on the single asyncio event-loop model and a smaller current-turn registry. | PR #60 is stronger for concurrent, multi-client server deployments; this PR is easier to reason about in the intended stdio/process-local usage but does not provide the same hardening. |
| Question requests | Does not support the newer Cobuild question-response API. | Supports retained question results and \`answer_cobuild_question\`, including answer constraints. | This PR covers the current SDK interaction model; PR #60 would need additional lifecycle and retention logic for questions. |
| Result-observation rule | Marks a turn observed once a turn ID/status is delivered, even if the terminal result is later read from retained history. | Requires the terminal result itself to be retrieved before another operation can start. | PR #60 favors throughput and recovery from retained history; this PR is more conservative, ensuring an agent cannot continue past an unseen terminal result. |

Neither approach is universally better. PR #60 is appropriate if the server must be hardened as a long-lived, multi-client service with strict memory limits, historical recovery, and adversarial concurrency concerns. This PR favors a smaller contract for the current process-local/stdio model: one conversation, one current turn, explicit inspection before continuation, no private confirmation-token API, and no automatic mutation of Cobuild state to enforce payload limits.


crmapj:
Implemented the lightweight recovery fix in `7edcd15`.

### What changed

- Exception-backed terminal turns now include `is_confirmation_request=false` and `is_question_request=false`, matching the shape that the next-turn checks already require.
- Added a regression test for the full failure lifecycle: SDK connection exception → retained status retrieval → successful next send.

### Why this shape

Live testing against the Chris personal DSS instance showed that normal Cobuild error responses already carry both flags and allow a later turn, while a transport exception omitted them and caused the next send to crash with `KeyError("is_confirmation_request")`.

The fix repairs that result contract at its source. It adds no fallback lookup, helper, state, lock, executor policy, or remote-server concurrency machinery, which keeps the implementation aligned with the current local, process-scoped MCP design.

### Validation

- Focused regression: passed
- Full test suite: `126 passed`
- Full pre-commit suite: passed
- Adversarial review: verified retained polling and next-send behavior, including the direct-indexing invariant
- Thermonuclear maintainability review: no structural regression or missed simplification; production change is two explicit fields and the module remains under 500 lines

pmasiphelps:
1. I added one validation for answer_cobuild_question: when rejected = true, answers list should be empty
2. I tightened the cobuild tool docstrings to fit the style of the rest of the tools. All of this guidance is already in the skills

crmapj (approved):
LGTM

PR #79, opened 2026-07-31
feat(skills): add Excel migration source subskill

pportal2:
## Summary

This splits migration guidance into a generic migration skill and an Excel-specific source subskill.

- Keeps shared planning, visual-only build, validation, and documentation/cleanup requirements in the generic skill.
- Adds Excel-specific workbook inspection, semantics, ingestion, parity validation, and helper scripts.
- Incorporates SME feedback: make projects independently reviewable, preserve sheet/table correspondence for inputs, and map each reviewable output sheet to its own terminal dataset.
- Delivers an overview webapp from a fixed template because unconstrained Cobuild generation did not reliably produce working webapps.
- Documents the observed 30-minute Cobuild timeout so Headless splits large tasks into coherent turns.

## Comparison runs

| Phase | [Headless + Excel migration skill](https://agent-dev-kit-testing.se-platform.dataiku-sandbox.io/projects/MIG_1651_NEW_BUSINESS_PLAN/flow/)<br>Fable 5 high + Cobuild GPT-5.5 | [CLI + migration skills](https://agent-dev-kit-testing.se-platform.dataiku-sandbox.io/projects/202607291648_NEW_BUSINESS_FORECAST/flow/)<br>Fable 5 high |
|---|---:|---:|
| Planning and analysis | ~13 min | ~13 min |
| Building the flow | ~73 min<br>10 turns + 1 repair | ~24 min |
| Validation | ~4 min | ~4 min |
| Webapp delivery | ~19 min<br>1 turn | ~7 min |
| Documentation and organization | ~15 min<br>3 turns | ~7 min |
| **Total** | **~2 h 05**<br>14 turns + 1 repair | **~55 min** |

Other Headless runs:
https://agent-dev-kit-testing.se-platform.dataiku-sandbox.io/projects/NVIDIA_DCF_20260730_1121/flow/
https://agent-dev-kit-testing.se-platform.dataiku-sandbox.io/projects/DEMOGRAPHICS_ACN_20260729_1813/flow/
https://agent-dev-kit-testing.se-platform.dataiku-sandbox.io/projects/202607311447_DEMOGRAPHICS_ACN/flow/

Source files:
https://dl.dataiku.com/file/poNj6w3jp4YrJjQn/5Txsj10xHvUwPArI/main%20(1).xlsx
https://dl.dataiku.com/file/poNj6w3jp4YrJjQn/JywVBlZFFcpWsDed/Demographics_ACN.xls
https://dl.dataiku.com/file/poNj6w3jp4YrJjQn/ZNg7weRqbpcXxvk8/Nvidia%20DCF.xlsx

## Contributing note

The Excel references follow the source-specificity and per-file size guidance in `CONTRIBUTING.md`. The deliberate exception is splitting the material across several focused Markdown files: the parsing, semantics, build, validation, and webapp contracts do not fit usefully into one file under the 90-line guideline.

## Not included yet

- Automatically enable Flow Zone descriptions through `flowDisplaySettings.showFlowZoneDescriptions`.
- Redirect users migrating sources other than Excel to the Dataiku Migrate initiative.


skills/dataiku-headless/references/cobuild.md
```diff
@@ -39,10 +39,10 @@ Do not use this guide when:
 5. Send the grounded request with `conversation_id` and `project_key`. Set `allow_edit_project=false` for inspection or explanation and `true` for an explicitly requested creation or modification.
 6. Retain the returned `conversation_id` for follow-up work.
 7. If Cobuild returns a delete confirmation request, inspect the deletion details and respond through `answer_cobuild_confirmation`.
-8. If Cobuild returns a question (`is_question_request=true`), read the `question` text and `predefined_answers`, then respond through `answer_cobuild_question`. Pass `answers` (one or more of the predefined answers, or a free-text answer when `allow_custom_answer=true` with `used_custom_answer=true`), or `rejected=true` to decline answering.
 
 ## Prompt Guidance
 
+- Keep each turn to one coherent unit that fits the server-side time cap, observed at 30 minutes; split wide or repetitive work across turns.
```
pmasiphelps: @crmapj @nmartorell does this line align with our recent cobuild reimplementation?

nmartorell: @pmasiphelps sort of, but even so I'm inclined to not include such a comment.

How the current Cobuild conversation implementation works:
- Headless uses a tool to start or obtain a conversation with Cobuild
- It then uses another tool to start a "turn" (message/deletion confirmation/qurestion answer) in a separate asyncio Task; after starting the Task, it waits up to 240s for the Task to finish (240s was chosen as Codex has a default 300s timout for tool calls)
- If the Task does not finish in 240s, then the "turn" tool call returns, instructing the harness to call the "wait_for_cobuild_turn" tool (which again waits for 240s, returns if not complete, and so on).

This is the "Cobuild conversation" lifecycle in Headless, which is what is currently described in the Cobuild SKILL. 

The Cobuild lifecycle (on the DSS side), which is currently not described in the SKILL, is different; in this case, we know that Cobuild will stop working (i.e. timeout) after 30min of work. 

Personally, I'm not sure we should document this on the Headless side. At the moment, if Cobuild times out during the execution of instructions, the harness will see this, inspect the project to see what Cobuild actually did (if anything), and then re-send new instructions.

The problem I see with the proposed addition to cobuild.md is that the harness using Headless really has not idea of knowing how long Cobuild will take to process an instruction. Thus, this instruction is either not useful, or potentially harmful (e.g. the harness could take it to the extreme and start sending cobuild instructions to build recipe-by-recipe, significantly increasing execution time). Curious @pportal2 if you saw any of this behavior during testing?

If we wanted to add anything, I would add a sentence that briefly explains the Cobuild lifecycle (as above), and then explicitly tells the harness how to correct for this (i.e. inspect what has been done, and re-issue instructions based on that). I would personally not add any instructions that try to influence the granularity of instructions sent to Cobuild (or rather, I would only do this if there are reasons beyond the Cobuild timeout to do this).

Just my 2 cents!

skills/dataiku-headless/references/migrations.md
```diff
@@ -27,32 +27,13 @@ The purpose of this phase is to inspect the project to be migrated (i.e. the `so
...
-    1. Create a Project Wiki.
-    2. Populate the Wiki with the migration plan created in Phase 1; it should be human-readable.
-    3. Populate the Wiki with the validation plan created in Phase 1; it should be human-readable and note the results of all validation checks.
-
-  - Cleanup
-
-    1. Delete any orphaned migration-created Flow assets (for example datasets and recipes) that are not part of the final migrated Flow.
-    2. Do not delete pre-existing user/project assets unless the user explicitly requests it.
-
+5. Write a Documentation and Cleanup Plan to `<bundle_dir>/migration_v<n>/documentation_and_cleanup_plan.md` that covers every Phase 4 requirement and distinguishes migration-created assets from pre-existing project assets.
```
nmartorell: @pportal2 given that the instructions for the `documentation_and_cleanup_plan.md` document have been removed, what do we expect harness going to include here? What's the motivation for removing them?

skills/dataiku-headless/references/migrations.md
```diff
@@ -70,13 +51,16 @@ Validation must confirm that the requested final output dataset is produced by a
 
 Validation must also confirm that the completed migrated flow contains no code recipes unless the user explicitly requested code.
 
-Validation is not complete until the migration also satisfies the Documentation and Cleanup Plan.
-
 ## Phase 4: Document and Cleanup
 
-Read the Documentation and Cleanup plan from `<bundle_dir>/migration_v<n>/documentation_and_cleanup_plan.md`. Apply the Documentation and Cleanup Plan fully via Cobuild.
+Read the Documentation and Cleanup Plan from `<bundle_dir>/migration_v<n>/documentation_and_cleanup_plan.md` and apply it via Cobuild, except for read-back and project settings:
```
nmartorell, line 62: Similar to my comment above: it now seems that we've moved the instructions for  `documentation_and_cleanup_plan.md` to this section. I think this will be confusing to the agent.

skills/dataiku-headless/references/migrations.md
```diff
@@ -70,13 +51,16 @@ Validation must confirm that the requested final output dataset is produced by a
...
+
+- Zone every migration-created Flow asset by stage or functional area, renaming the undeletable default zone for the first stage. Enable `flowDisplaySettings.showFlowZoneDescriptions` through project settings or record it as blocked; Cobuild cannot change this setting.
```
nmartorell: A couple of things:
- I would not use the specific JSON keys for `flowDisplaySettings.showFlowZoneDescriptions`
- It is not clear how the agent should modify this setting: if we provide a direct mutating tool to do so, we should name the tool, otherwise, we should tell it to record this (and possibly remind the agent that Cobuild cannot mutate project settings).

skills/dataiku-headless/references/migrations.md
```diff
@@ -70,13 +51,16 @@ Validation must confirm that the requested final output dataset is produced by a
...
+- Set the project's short and long descriptions. For every migration-created dataset, recipe, and zone, set the field DSS displays: dataset `description`; recipe and zone `shortDesc` (ask Cobuild for "short description"; their long `description` is not rendered). Rename generated `compute_<output>` recipes for the transformation they perform.
```
nmartorell: Here I think the JSON keys are appropriate, as there are indeed multiple description-like fields in some objects.

skills/dataiku-headless/references/migrations.md
```diff
@@ -70,13 +51,16 @@ Validation must confirm that the requested final output dataset is produced by a
...
+- Create a human-readable Project Wiki containing the migration plan, validation plan and results, and a final-output column dictionary. Keep column documentation out of datasets because it drifts through downstream recipe schemas.
```
nmartorell: Similar to above - should this be included in the instructions for ``flowDisplaySettings.showFlowZoneDescriptions`?

skills/dataiku-headless/references/migrations.md
```diff
@@ -70,13 +51,16 @@ Validation must confirm that the requested final output dataset is produced by a
...
+- Apply the cleanup safety rule below. Create and run a rebuild scenario covering every final output; its job result is the final build proof.
 
-The migration is not complete until all required documentation, Flow Zone, Wiki, and cleanup tasks from that plan have been completed.
+Use read tools to enumerate every zone, dataset, and recipe, verify each displayed description is non-empty (`short_description` for zones and recipes; `description` for datasets), and write the inventory to `<bundle_dir>/migration_v<n>/documentation_evidence.md`. Completion claims require this inventory, validation evidence, and the scenario job result; never rely on the Cobuild report.
```
nmartorell, line 68: I like this here - it provides additional "smack on the head" verification instructions to the agent, in case it ignores the content of ``flowDisplaySettings.showFlowZoneDescriptions`.

skills/dataiku-headless/references/migrations.md
```diff
@@ -87,6 +71,8 @@ Unless the user explicitly requests a code-based transformation, migrations must
 
 Do not use Python, SQL, R, or other code recipes merely because the logic is awkward, stateful, easier to express in code, or difficult to reproduce visually. Difficulty is not an exception.
 
+A statistic missing from a visual recipe's aggregate list is not yet a justification either: quantiles, medians, and modes compose from Window ranking and TopN recipes. Record a deviation only after the visual composition genuinely fails.
```
nmartorell, line 79: @pmasiphelps generic enough or overfitting?

skills/dataiku-headless/references/migrations.md
```diff
@@ -103,6 +91,10 @@ The requested final output dataset must be produced in DSS from the migrated sou
 
 Source-boundary fidelity alone is not sufficient: the migrated project must contain the DSS flow that performs the transformation.
 
+## Gap resolution
```
nmartorell, line 99: I'm not sure I understand this section, what is it supposed to do? I know we mention `needs-human-input` above, but I'm still not sure where this should be "recorded" and what should be done about it.

skills/dataiku-headless/references/cobuild.md
pportal2: @nmartorell I did see Cobuild hit the timeout several times during Headless runs.

In the context of building a Dataiku flow end-to-end, doing it recipe by recipe is actually the expected behavior and does not necessarily increase execution time. Headless asks Cobuild to create a recipe, runs it, and checks the result before continuing. If several dependent recipes are created in one turn and the first one is incorrect, the downstream recipes may also need to be reworked after fixing the upstream logic, which can cost more time overall.

The intent of the guidance is to avoid sending very wide or repetitive tasks that could cause Cobuild to time out mid-turn. In practice, this change made a real impact as I haven't seen any timeouts since.

Regarding the 30-minute limit specifically, my thinking was that Headless might be able to infer that the timeout was approaching after several polling cycles and act accordingly. However, I agree that it may not have reliable awareness of the elapsed Cobuild processing time, and I haven't verified this behavior in practice yet.

nmartorell (commented):
Hi @pportal2, I left a few comments on the migration and cobuild md files. In addition:

- Al instances of "DSS" should probably be replaced for "Dataiku" (as per PR #82 )
- I would recommend the structure of the py and skills docs as:

```
  sources/excel/
    excel.md                 # required entry point / routing
    guides/
      build.md
      reading-workbooks.md
      excel-semantics.md
      validate.md
      webapp.md
    helpers/
      dump_anatomy.py
      dump_workbook.py
    webapp/
      template.*
```

Finally, I only did a very light review of the webapp part of the PR. I would like to do some functional testing, do you have any Excel workbooks I could use to test this out?

skills/dataiku-headless/references/cobuild.md
nmartorell: When there was a Cobuild timeout, did Headless recover semi-gracefully? e.g. did it inspect created artifacts, and then re-instruct Cobuild to create the remaining ones?

In any case, if you didn't see significant increases in execution time from a recipe-by-recipe vs full-flow approach, then I would structure this as follows:

- cobuild.md: make the agent aware the Cobuild may time out when asked to perform large transformations; if it does, it should inspect the project, and re-issue instructions to continue working on what it hadn't finished. 
- migration.md: strong opinions go here; we should tell it to create the plan, but then migrate in a recipe-by-recipe. 

wdyt?

Overall though - what is the objective in attemtping to avoid Cobuild timeouts? Do we see worse performance when there are timeouts, increated token usage, time, etc.? Or is it more that you'd like to avoid the timeouts because they look untidy in the Cobuild UI? (Which would be a valid reason imo, just trying to understand).

To provide more context on cobuild.md: I don't think we should include any optinionated "break down cobuild instructions into X/Y/Z", because in interactive use the agent should do whatever the user asks it to. i.e. if the users wants to one-shot huge projects.... well then that's on them. The one thing I find a bit annoying about Cobuild is that I sometimes ask it to do X, and then it replies with "I did the first half of X, should I continue?", which is a behavior I'd like to avoid here.

pportal2:
@nmartorell See PR text, I have provided 3 excel files to test this out

skills/dataiku-headless/references/cobuild.md
pportal2: I agree with that split:
- cobuild.md: explain that large transformations may time out and, when they do, Headless should inspect the artifacts already created and re-issue instructions only for the remaining work.
- migration.md: recommend executing the migration as coherent, independently verifiable units.

I did not inspect the recovery sequence closely, but Headless managed to continue after the timeouts.

I think avoiding timeouts has value beyond keeping the UI tidy. The principle of delegating to a subagent is to give it the right context for a bounded, reviewable unit of work, then inspect and verify the result before continuing.

This also feels more effective and likely less costly. With one large prompt, Cobuild may consume a large context and many tokens before timing out. Headless must then inspect the partial result, determine exactly where Cobuild stopped, reconstruct the necessary context, and prompt it again.

With bounded units of work, the workflow is easier to manage: Headless asks Cobuild to complete one focused task, inspects and verifies the result, and then moves to the next unit. The heavier planning remains with the stronger orchestrator model, while Cobuild focuses on execution. This approach may be quite specific to migration workflows, so I agree that the split you proposed is the right one.

pportal2:
Hi @nmartorell, thanks for the feedback. 

Here’s my take on some of the points:

- flowDisplaySettings.showFlowZoneDescriptions: I agree that it is mentioned too often. I’ll remove the repeated references and keep a single instruction explaining that, for now, the user must enable flow zone descriptions manually.

- documentation_and_cleanup_plan.md instructions: I moved the plan generation instructions to Phase 4 and expanded them with requirements that I think materially improve project quality: functional flow zone, renaming the default zone, short descriptions, meaningful recipe names, a rebuild scenario, etc... I haven’t observed the agent getting confused in practice, but I agree that the instructions for writing the plan should be located where the plan is initially created. I’ll move them back while keeping the execution and verification guidance in Phase 4.

- Excel source structure: I agree with the structure you suggested. I originally had something similar, but changed it to comply with the [helper script rule](https://github.com/dataiku/dataiku-headless/blob/main/skills/dataiku-headless/references/migrations/CONTRIBUTING.md#helper-scripts-optional) in migrations/CONTRIBUTING.md. Should we remove this file to avoid encouraging additional migration sources, or revise it to fit the new architecture?

skills/dataiku-headless/references/migrations.md
pportal2: It forces the agent to handle certain types of source data correctly:
- derivable means it does not need to be a source dataset because it can be derived from another source. Create a recipe that calculates it instead of uploading it as a dataset, which was an observed behavior.
- hand-authored means the data was edited manually. Preserve it as a source dataset instead of trying (and failing) to derive it from other data, which was also observed.
- needs-human-input means it is unclear how the source data should be handled. Force it to ask the user, not to guess (also observed behavior).

This instruction was enough to guide the agent, but since these terms are not defined elsewhere, we could clarify them:
```
## Gap Resolution

Classify every non-obvious construct:

- 'derivable': a function of its inputs, such as formula chains, lookups, or query steps. Migrate it as Flow logic.

- 'hand-authored': manually entered or edited values with no recoverable rule, such as overrides, pasted reference data, or inconsistently cleaned columns. Preserve the source values as data, document them, and never re-derive them.

- 'needs-human-input': answerable only by the workflow owner, such as value provenance, redaction artifacts, contradictory configuration, or business-nonsensical wiring. Ask; do not guess.

Carry unresolved items in the Migration Plan and surface them at the Phase 1 step 6 gate.
```

nmartorell:
Hey @pportal2 ,

Thanks for taking the time to read through the comments, I agree with the changes. I think the only thing we have left to work out is the prompt guidance text.

Maybe something like:
```
Unless the user explicitly requests otherwise, keep each turn to one coherent unit of work. For example, creating a recipe, adding descriptions to a number of objects, editing a Wiki. The intention here is to not submit large, monolithic instructions to Cobuild that cover many actions at once.
```

pportal2:
@nmartorell @pmasiphelps 

Feedback integrated:
- DSS → Dataiku wording across the subskill (per #82)
- Cobuild turn-granularity guidance split between `cobuild.md` and `migrations.md`
- Documentation & Cleanup Plan content requirements back in Phase 1; Phase 4 keeps execution + verification
- Dropped the raw `flowDisplaySettings` JSON key, plain-language setting instead
- Gap resolution expanded with `derivable` / `hand-authored` / `needs-human-input` definitions
- `sources/excel/` restructured into `excel.md` + `guides/` + `helpers/` + `webapp/`; removed `CONTRIBUTING.md`

nmartorell (approved):
I agree with the changes - looks good to merge.

PR #80, opened 2026-07-31
feat(onboarding): test instance connection

yjouini:
## Summary

- add a non-persisting Test connection action to the local onboarding page
- verify the entered credentials with DSS before saving
- preserve the entered values and display an inline result

## Validation

- uv run pytest -q (136 passed)
- ruff check dataiku_mcp/setup_server.py
- setup-page smoke


pmasiphelps (approved):
looks great.

PR #83, opened 2026-07-31
feat(onboarding): require a tested connection before saving

yjouini:
## Summary
- place the certificate-verification bypass under folded Advanced options
- require a successful connection test before an instance can be saved
- invalidate the validation when the URL, API key, or certificate setting changes
- make the disabled Save state visibly greyed out

## Validation
- `.venv/bin/python -m pytest` (153 passed)
- `.venv/bin/ruff check .`
- browser smoke of the configuration page and folded Advanced options


pmasiphelps (approved):
looks great. i just added one small thing to preserve local filled values if the form connection does not pass validation

PR #84, opened 2026-08-02
doc(skills): document embedded visual recipe stages

pportal2:
Adds a compact capability map for visual data-prep recipes, covering embedded pre/post-filters, computed columns, custom or global aggregations, and output controls. This gives Headless the recipe-level knowledge needed to produce leaner migration flows with fewer helper recipes and datasets, while preserving meaningful intermediates.


nmartorell:
Hi @pportal2, I think this is a very very useful addition to the visual recipe skills doc (it is information that was unfortunately lost when moving to the "Cobuild mode" of Headless).

I have a few commends on the specifics of the implementation:
- The "Recipe Types | Embedded Transformations" table includes specific DSS JSON keys, which is something that we've been avoiding now that we've moved to delegating recipe payload generation to Cobuild; I can see us wanting to make an exception if the JSON payload keys were particularly nondescript or misleading, but I think it this case they are quite explanatory.
- Line 33 (i.e. on consolidating pre/post transformations into a single recipe) is too strong; there are other cases (e.g. observability, maintainability, etc.) in which having separate, individual recipes would make sense.
- Line 35 (i.e. on Cobuild building instructions) seems unnecessary; these are mutating Cobuild instructions.

I ideated a bit with Codex, and thought the following structure might make sense. It keeps the same information (minus the Cobuild sentence) as you have, with the following differences: (a) removes specific API payload key names, (b) softens the instructions on using pre/post recipe transformations, and (c) places the transformation descriptions in their own table (which seems a bit clearer to me).

In any case though, you've spent a lot more time than me testing Migrations and visual recipe creation, so totally fine if you don't agree with these comment ;) 

```
## Embedded Recipe Stages

Several visual recipes contain processing stages around their named action. Inspect the full `payload` from `get_recipe_settings`; do not assume that one Flow node contains only the core action.

| Stage | Purpose and effect |
| --- | --- |
| Input pre-filter | Removes rows before the core action. It can change join inputs, group membership, aggregate denominators, or window partitions. |
| Pre-action computed columns | Derives or changes columns available to the core action. |
| Core action | Performs the recipe's named operation, such as joining, grouping, ranking, or splitting. |
| Post-action computed columns | Derives columns from the action result. For joins, these can use columns from both inputs. |
| Post-filter | Removes rows after the core action and can filter generated aggregate or joined columns. |
| Output controls | Select, rename, map, or otherwise shape delivered output columns. |

Do not move a condition between stages unless it is equivalent at both grains.

| Recipe types | Input pre-filter | Pre-action computed columns | Core action and other features | Post-action computed columns | Post-filter | Output controls |
| --- | --- | --- | --- | --- | --- | --- |
| `grouping` | Yes | Yes | Grouping keys; standard or SQL-engine custom aggregations | — | Yes | Selection and name overrides |
| `window` | Yes | Yes | Window definitions and values; window-wide global aggregations | — | Yes | Retrieved-column selection |
| `join`, `fuzzyjoin` | Per input | Per input | Join action | Yes | Yes | Selected columns |
| `sort`, `distinct`, `topn` | Yes | Yes | Ordering, deduplication, ranking, or row limits | — | Yes | Yes |
| `vstack` | Per input | Yes | Schema mapping and optional origin column | — | Yes | — |
| `split` | Yes | Yes | Split conditions and catch-all output | — | Yes | — |

Prefer one recipe when its embedded capabilities express the transformation and separating it provides no material operational, performance, governance, or reuse benefit.
```

Note: the first sentence is unchanged.

pportal2:
I agree with the changes @nmartorell, please commit to the PR branch

nmartorell (approved):

PR #86, opened 2026-08-03
feat(code-envs): add code environment administration

nmartorell:
## Summary
- add Python and R code environment create, update, and delete tools
- enrich code environment listing with exact search, details, and bounded usage inspection
- add skill guidance and isolated tests

## Validation
- ruff check
- pytest (94 passed)
- Python syntax compilation

Trello: https://trello.com/c/PAtoALkG/118-create-update-delete-code-env-tools-and-skills


pmasiphelps (approved):
this looks great! my only change was to update language such that only `DESIGN_MANAGED` code envs can accept package changes via the `requested_packages` arg, and other types of code envs can still accept other updated (e.g. delete, rebuild env, rebuild images).

PR #87, opened 2026-08-04
refactor(cli): use one direct server entry point

pmasiphelps:
## Summary

- remove the unused `serve` subcommand and the duplicate `dataiku-mcp` command
- keep `dataiku-headless` as the only command
- remove Typer and the dependencies it brought in
- move the small command-line wrapper into `dataiku_mcp/cli.py`, next to the rest of the application code
- support normal `--help` and `--version` output, and show an error for unknown arguments

Running `dataiku-headless` with no arguments still starts the MCP server. Plugin startup is unchanged.

## Validation

- 158 tests pass
- pre-commit checks pass
- the package builds successfully
- the installed command was tested from the built wheel


nmartorell:
I restored the direct module entry point (python -m dataiku_mcp) because routing it through the new CLI eagerly looks up installed package metadata. That lookup raises PackageNotFoundError when the server is run from a source checkout that has dependencies available but has not been installed as a distribution. The installed dataiku-headless console command still uses the CLI and retains its --help and --version behavior. I also added a regression test to ensure the module entry point calls the server directly without loading the distribution-backed CLI.

pmasiphelps:
@nmartorell good change - good to merge?

pmasiphelps:
@nmartorell i removed the cli altogether!

nmartorell:
@pmasiphelps this looks good to me. I also pushed a commit that removes the console entry point all together, i.e. `uv run dataiku-headless`. This leaves the following entrypoints to the project:

- Run as a harness plugin (via launcher.sh)
- Run directly as an mcp server (via launcher.sh or run_mcp.py)
- Run by starting the harness in the cwd of the checked-out gh repo, which picks up .mcp.json

I think it might also be worth discussing the removal of __main__.py, to also remove the `python -m dataiku_mcp` entrypoint, but we can discuss in standup pros/cons and decide then.

The current PR lgtm for a merge :)

nmartorell (approved):

PR #90, opened 2026-08-06
feat(admin): allow non-admin user and group listing

nmartorell:
## Summary
- allow user listing to fall back to the non-admin user-info API
- allow group listing to fall back to the non-admin group-info API
- document limited non-admin responses and cover both authorization paths

## Validation
- uv run pre-commit run --all-files --show-diff-on-failure
- uv run pytest
- uv build
- uv run cz check --rev-range origin/main..HEAD


pmasiphelps (approved):
looks great! i changed the formatting of one test to pass the ruff check, and i dropped the change to administration.md, because it will be superceded by the better wording coming from #86  "Use this routing guide for instance-level administrative tasks and the corresponding direct MCP tools. Permissions vary by operation; follow the matching guide."

PR #92, opened 2026-08-07
fix(codex): reference bundled MCP manifest

pmasiphelps:
This fixes the Codex plugin import error:

image (15)

By moving the mcp launcher script into .mcp.json instead of .codex-plugin/plugin.json

OpenAI ref doc: https://developers.openai.com/plugins/build/plugins


pmasiphelps:
We think this is an issue with the Dataiku ChatGPT enterprise plugin uploading flow - not with actual plugin distribution

nmartorell:
Already discussed with Pat, but noting here that this breaks the "auto-start" for Claude (when started in the checkout directory of the repo). But as Claude is the only harness that we currently support auto-start, dropping support is acceptable.

nmartorell (approved):

PR #94, opened 2026-08-10
fix(connections): redact prefixed secret fields

### LINKED ISSUE #93: fix(connections): clientSecret is not redacted by get_connection_info
`get_connection_info` returns OAuth application secrets in cleartext, violating the `AGENTS.md` rule: "Never return API keys, credentials, unredacted connection secrets, or sensitive exception text."

## Reproduction

1. Call `get_connection_info` on any connection whose params carry an OAuth app config (e.g. a Databricks connection with `authType: OAUTH2_APP`).
2. `params.clientSecret` and `resolvedParams.clientSecret` are returned verbatim.

Sibling fields in the same response (`password`, `resolvedOAuth2Credential`) are correctly redacted.

## Cause

`_is_sensitive_key` (`dataiku_mcp/tools/connections.py:39-53`) normalizes `clientSecret` to `clientsecret`, then:

| Check | Result |
|---|---|
| `normalized in _SENSITIVE_EXACT_KEYS` | miss — set holds `secret`, not `clientsecret` |
| `endswith` any of `password`, `privatekey`, `secretkey`, `sessiontoken` | miss — key ends `…secret`, not `…secretkey` |
| `endswith("credential")` | miss |
| `endswith("token")` | miss |

Returns `False`; `_redact_sensitive_data` passes the value through.

## Scope

Root cause is an inconsistency between exact-match and suffix-match words, so this generalizes beyond `clientSecret`:

| Word | Matched as | Prefixed variant redacted? |
|---|---|---|
| `password` | suffix | ✅ `clientPassword` |
| `privateKey` | suffix | ✅ `sshPrivateKey` |
| `secretKey` | suffix | ✅ `awsSecretKey` |
| `secret` | exact only | ❌ `clientSecret` |
| `apiKey` | exact only | ❌ `openaiApiKey` |
| `accessKey` | exact only | ❌ `awsAccessKey` |
| `credentials` | exact only | ❌ `userCredentials` |

Affects both call sites: `get_connection_info` (`connections.py:245`) and `test_connection` (`connections.py:267`). Not connection-type specific.

No test coverage — `_redact_sensitive_data`, `_is_sensitive_key`, and `__DATAIKU_REDACTED__` have no references under `tests/`. Logic unchanged since the initial commit.

## Impact

No privilege escalation: the caller already holds connection-read rights and the DSS API returns the secret. The harm is the egress this layer exists to prevent — secrets entering an LLM context and persisting in transcripts and logs.

## Proposed fix

Move `secret`, `apikey`, `accesskey`, and `credentials` from `_SENSITIVE_EXACT_KEYS` into `_SENSITIVE_SUFFIXES`; add tests for the prefixed variants.

Keep suffix semantics rather than substring — substring would swallow `tokenEndpoint` and `authorizationEndpoint`, which are non-secret and currently pass correctly.

Tradeoffs:
- `credentials` as a suffix also matches `storageCredentials` (a config list), hiding whether any are configured.
- Pre-existing, unrelated: `properties[].secret` is over-redacted; that field is a boolean marking whether a property is secret, not a secret value.

crmapj:
## Summary

- redact prefixed secret fields such as `clientSecret`, `openaiApiKey`, `awsAccessKey`, and `userCredentials`
- replace the conflicting exact-match and suffix-match rules with one suffix policy
- cover nested `params` and `resolvedParams` while keeping endpoint fields visible

## Root cause

The redactor treated several sensitive names as exact matches, so their prefixed variants passed through unchanged. Using one suffix policy removes that inconsistency and preserves all existing matches.

## Impact

Connection inspection and testing no longer return these secret families to the MCP client. This does not change the tool surface or Dataiku permissions.

## Validation

- `uv sync --locked`
- `uv run pytest` — 196 passed
- Python syntax check
- `uv build`
- adversarial maintainability review and representative old/new predicate comparison

Fixes #93


crmapj:
@pmasiphelps

pmasiphelps (approved):
looks great!

pmasiphelps:
@crmapj im seeing this in draft state - is it ready to go or are you still working on it?

crmapj:
@pmasiphelps ready to go on my side

PR #96, opened 2026-08-11
feat(plugins): support Windows uv launches

nmartorell:
## Summary
- Launch Codex and Claude MCP servers directly with uv instead of POSIX sh.
- Document uv 0.12.0+ as the runtime prerequisite and retain launcher.sh as inactive legacy fallback code.
- Refresh contributor and PEP 723 documentation to match the active launch path.

## Motivation
The plugin manifests previously required sh and bin/launcher.sh, preventing native Windows hosts from launching the stdio server. uv runs the existing PEP 723 entry point cross-platform, removing the Git Bash, WSL, Node, and separately provisioned Python requirement for Windows users.

## Validation
- Reviewed the diff against origin/main.
- Verified both manifest JSON command and argument definitions statically.
- Ran git diff --check.
- Tests, builds, and formatting were not run per request.


nmartorell:
@hanneshapke @pmasiphelps I tested with Codex on Windows and it worked well. Will test with Claude as well and confirm.

nmartorell:
@pmasiphelps tested the changes on Claude and Codex on Windows and macOS.

pmasiphelps (approved):
Tested Codex and Claude Code in the remote windows instance - both worked great!

PR #106, opened 2026-08-12
fix: pin Dataiku instances per tool request

### LINKED ISSUE #102: fix: pin the Dataiku instance for each tool invocation
## Problem

The active Dataiku instance is stored as mutable process-global state. Many tool handlers yield before resolving their Dataiku client, then call `get_dss_client()` inside a worker. A concurrent `switch_instance` request can therefore redirect an already-started operation to a different configured instance.

For example, `set_project_variables` awaits `ctx.info` before its worker resolves the client. The same pattern appears across read, write, and admin tools. Some admin tools can also verify permission against one instance and mutate another.

## Impact

A request intended for instance A can read from or mutate instance B, including project variables, uploads, user/group administration, and job operations.

## Suggested fix

Capture the `DSSInstance` or `DSSClient` before the first `await` in every tool invocation, then pass that captured client into all blocking workers and follow-up polling. Avoid resolving `get_dss_client()` from inside a deferred lambda. Add concurrency tests that switch instances while a tool is paused before its worker starts.

## Review evidence

- `dataiku_mcp/tools/utils/auth.py:50`
- `dataiku_mcp/tools/projects.py:140`
- `dataiku_mcp/tools/instances.py:38`

mjeckmans:
Fixes #102.

This change snapshots the active Dataiku instance when an MCP tool starts and propagates that request context into both blocking executors. A later `switch_instance` call can no longer redirect queued work, including admin checks and follow-up polling.

Tests and checks run:

- `uv run pytest tests/test_instance_pinning.py tests/test_project_variables.py tests/test_cobuild.py`
- `PYTHONPYCACHEPREFIX=/tmp/dku-pycache uv run python -m py_compile dataiku_mcp/__init__.py dataiku_mcp/config.py dataiku_mcp/tools/utils/async_executor.py tests/test_instance_pinning.py`
- `uv run pytest`
- `uv build`


mjeckmans:

nmartorell:
Excellent catch @mjeckmans - I modified the implementation slightly (to use FastMCP middleware instead of using a DataikuMCP class), but overall the approach is sound.

@yjouini would you mind reviewing before merging, as we both worked on the instance selection logic, so would appreciate a second pair of eyes. Thanks :)

yjouini (approved):
Nice catch ! LGTM

PR #107, opened 2026-08-12
fix: preserve existing upload datasets

### LINKED ISSUE #103: fix: preserve an existing dataset when replacement upload fails
## Problem

When `overwrite=True`, `_create_uploaded_dataset_from_file` deletes the existing dataset before it opens or validates the supplied local file. It creates the replacement only afterwards.

## Impact

A missing or unreadable input file, or a later upload/autodetect failure, leaves the original dataset deleted and can leave a partial replacement behind.

## Suggested fix

Validate that the input is a regular, readable file before any remote mutation. Where the API permits it, stage the replacement under a temporary dataset and replace the original only after upload and schema detection succeed. At minimum, return a recovery-oriented error if a post-delete step fails.

## Acceptance criteria

- An invalid local path with `overwrite=True` does not change an existing dataset.
- Tests cover unreadable files and failures during upload/autodetection.

## Review evidence

- `dataiku_mcp/tools/datasets.py:22`
- Existing dataset deletion occurs at line 40, before local-file open at line 43.

mjeckmans:
Fixes #103.

Uploaded Files dataset creation is now idempotent and fail-closed: a new target is created once, but an existing target is never deleted or replaced. The Dataiku client exposes no atomic replacement operation, so `overwrite=true` now returns a clear error without changing the existing dataset. Local input is opened before a new remote dataset is created.

Tests and checks run:

- `uv run pytest tests/test_datasets.py tests/test_tool_surface.py`
- `PYTHONPYCACHEPREFIX=/tmp/dku-pycache uv run python -m py_compile dataiku_mcp/tools/datasets.py tests/test_datasets.py`
- `uv run pytest`
- `uv build`


mjeckmans:

pmasiphelps:
@mjeckmans please take a look at my changes:

- Restored overwrite=true to delete and recreate the existing Uploaded Files dataset.
- Kept file validation before deletion, preventing a missing local file from deleting anything.
- Documented overwrite as explicit, destructive, and non-atomic.

mjeckmans:
Thanks for the follow-up. I think these changes are directionally reasonable and preserve backward compatibility for callers that rely on `overwrite=true`.

They fix the confirmed data-loss case for a missing or unreadable local file by opening the file before deleting the existing dataset. However, the broader failure window remains: after the delete, `create_upload_dataset`, `uploaded_add_file`, `autodetect_settings`, `settings.save`, or `get_schema` can still fail, leaving the original dataset gone and possibly leaving a partial replacement.

That is the second half of issue #103, and the current tests do not cover it—they cover invalid local input and successful replacement, but not upload or schema/autodetection failures after deletion.

I would suggest adding:

- failure-path tests for upload and autodetection/schema handling;
- a recovery-oriented error stating that the original dataset was deleted and replacement failed;
- cleanup or clear reporting of a partially created replacement where possible.

The ideal implementation would stage the new upload under a temporary dataset, validate it fully, and then replace the original. If the Dataiku API cannot atomically rename or swap datasets, documenting the behavior as destructive and non-atomic is appropriate, but the failure should still be explicit and actionable.

I would characterize this as fixing the invalid-path data-loss bug while leaving the post-delete failure window unresolved—not a reason to reject the compatibility change outright, but worth addressing before considering issue #103 fully closed.

pmasiphelps:
@mjeckmans how about if we just dont allow overwriting?

pmasiphelps:
@nmartorell can you take a peek at this? removing the overwrite option for creating uploaded file datasets

nmartorell (approved):
@pmasiphelps added a small change to make the recvoery path via cobuild clear. But feel free to revert if you don't like it! Either way, good to merge imo.

PR #108, opened 2026-08-12
fix: escape formulas in CSV exports

### LINKED ISSUE #104: fix: escape spreadsheet formulas in CSV exports
## Problem

`export_dataset` writes values directly to CSV cells. Spreadsheet applications can interpret cells beginning with `=`, `+`, `-`, `@`, tab, or carriage return as formulas when the resulting file is opened.

## Impact

Untrusted values stored in a Dataiku dataset can become spreadsheet formula injection when an exported CSV is opened in Excel or LibreOffice.

## Suggested fix

Escape formula-leading text values before passing them to `csv.writer`, for example by prefixing a single quote. Account for leading tab/carriage-return characters that spreadsheet applications may trim. Document the escaping behavior and add regression tests for every dangerous prefix.

## Review evidence

- `dataiku_mcp/tools/datasets.py:260`
- Row values are written without sanitization at lines 322-326.

mjeckmans:
Fixes #104.

CSV exports now prefix formula-like text with an apostrophe before writing it. This prevents spreadsheet applications from evaluating values beginning with formula prefixes while preserving ordinary strings and numeric values.

Tests and checks run:

- `uv run pytest tests/test_datasets.py`
- `PYTHONPYCACHEPREFIX=/tmp/dku-pycache uv run python -m py_compile dataiku_mcp/tools/datasets.py tests/test_datasets.py`
- `uv run pytest`
- `uv build`


mjeckmans:

pmasiphelps (approved):
@mjeckmans looks good! i made a couple changes:

1. consolidated helper functions from 2->1
2. escaped column names too

im approving the PR. please take a look at my changes. once you approve, we can merge

PR #109, opened 2026-08-12
fix: bound project library regex searches

### LINKED ISSUE #105: fix: bound regex project-library searches
## Problem

`search_project_library` accepts arbitrary Python regular expressions and evaluates them against every line of every readable library file. There is no pattern, input-size, file-count, or execution-time bound. `max_matches` only stops after a match is found.

## Impact

A catastrophic-backtracking pattern such as `(a+)+$` against a long non-matching line can monopolize a shared executor worker. Repeated requests can exhaust the worker pool and degrade unrelated MCP tools.

## Suggested fix

Use a timeout-capable regex engine, or bound searched file/line sizes and reject patterns that exceed a defined complexity policy. Add a total scan-byte/file limit independent of `max_matches`, and return an explicit truncation result when that limit is reached.

## Review evidence

- `dataiku_mcp/tools/project_libraries.py:343`
- User-controlled regex compilation is at lines 364-371.
- It reads each file and evaluates every line at lines 375-406.

mjeckmans:
Fixes #105.

Project-library regex search now uses the timeout-capable `regex` engine with a 100 ms limit per line. Pathological expressions return an actionable timeout error instead of monopolizing an MCP executor worker; ordinary regex and substring search behavior is preserved.

Tests and checks run:

- `uv run pytest tests/test_project_libraries.py tests/test_pep723_launcher.py`
- `PYTHONPYCACHEPREFIX=/tmp/dku-pycache uv run python -m py_compile dataiku_mcp/tools/project_libraries.py tests/test_project_libraries.py`
- `uv run pytest`
- `uv build`


mjeckmans:

pmasiphelps:
@mjeckmans i like the intent of this PR. i made a small commit to make the code slightly more readable IMO.

my main concern is adding the regex python package dependency. right now, we only require 3 packages: fastmcp to run the MCP server, python-dotenv to load env variables for credential overrides, and dataiku-api-client. the regex pypi package seems to have just one maintainer - is this legit? is resolving this issue worth this additional requirement?

mjeckmans:
Thanks for raising this. I agree that adding a native third-party dependency deserves scrutiny, especially given the project currently has a very small dependency set.

The reason for `regex` is the execution-time guarantee: this MCP endpoint accepts user-supplied regular expressions and searches project-library content line by line. Python’s stdlib `re` has no supported per-search timeout or cancellation API. Wrapping `re.search()` in a thread/future would only stop waiting; it would not stop a running catastrophic-backtracking match, so the executor worker could remain occupied. Signals are platform/thread constrained, and modifying private `re` internals would be CPython-version-dependent.

The `regex` package is specifically used because it provides a timeout parameter while keeping an API close to `re`. It is a legitimate production/stable PyPI package with verified maintainer metadata, recent releases, and documented upstream timeout support:

- https://pypi.org/project/regex/
- https://github.com/mrabarnett/mrab-regex/blob/hg/changelog.txt
- https://docs.python.org/3.10/library/re.html

I think the dependency is justified for this safety property, but I agree the PR should make the tradeoff explicit. I would add an end-to-end timeout test through `search_project_library`, verify the supported platform/wheel matrix, and document why stdlib `re` cannot provide the same guarantee.

A dependency-free alternative would be running stdlib `re` in a separate killable process, but that adds process/IPC overhead and considerably more lifecycle complexity for a line-by-line search. Pattern heuristics or input-length limits would reduce risk but would not provide a reliable execution bound.

mjeckmans:
One additional risk assessment: I would classify this as a **medium availability risk**, rather than a data-integrity or confidentiality issue.

Project-library searches run in the general thread pool, which defaults to 4 workers. A pathological stdlib `re` expression can occupy one worker for a long time; four concurrent searches could delay unrelated MCP operations. Timing out the async wait would not terminate the underlying regex execution, so the worker would remain occupied until the match completes. `max_matches` does not protect against a pathological nonmatching line, because the limit is only reached after matches are found.

The practical risk is lower in a trusted single-user deployment where project libraries are small and callers are known. It becomes materially higher when the MCP server is shared by multiple users/agents or exposed to automated, user-supplied regexes. In that setting, this is worthwhile availability hardening before broad deployment, although I would not characterize it as a critical vulnerability.

If we decide not to add `regex`, conservative pattern/input limits could be an interim mitigation, but they would not provide a strict execution-time guarantee. My preference is to retain the focused dependency for the timeout guarantee, subject to the additional end-to-end and platform validation mentioned above.

pmasiphelps:
closing because the risk seems low given we only support local stdio mcp server, and the additional dependency isnt worth it

PR #101, opened 2026-08-12
fix: harden Cobuild blocking interactions

mjeckmans:
## Summary

- retain terminal Cobuild turn results explicitly so pending actions remain discoverable
- expose the exact pending question or confirmation through `get_cobuild_pending_action`
- include pending action details in `list_cobuild_conversations`
- keep message and answer flows bound to the exact current `turn_id`
- document the required stop, inspect, and answer workflow
- add regression coverage for pending questions, confirmations, listings, and failed turns

## Validation

- `uv run pytest tests/test_cobuild.py tests/test_tool_surface.py` — 47 passed
- Ruff — passed
- `git diff --check` — passed

This PR intentionally contains no plugin changes; those overlap with PR #88.

PR #110, opened 2026-08-13
docs(cobuild): clarify cross-channel action recovery

nmartorell:
## Summary

- Document that Cobuild conversations can be continued through the Dataiku UI and MCP/API concurrently.
- Explain that questions and deletion confirmations are single DSS-side actions: the first channel to answer consumes the action.
- Document the safe recovery path for `No pending question/confirmation found`: do not retry, inspect project or UI state, then continue the same conversation if appropriate.
- Clarify `get_cobuild_turn_status` behavior and reuse common current-turn validation internally.

## Motivation

Live testing reproduced a UI answer consuming a pending Cobuild action created through MCP. A subsequent MCP answer correctly receives DSS's `No pending question found`, while the underlying conversation remains usable for a follow-up message. This PR records that behavior for operators and keeps the MCP surface unchanged.

## Relationship to #101

[#101](https://github.com/dataiku/dataiku-headless/pull/101) helpfully identified the importance of retained-turn recovery and prompted the investigation. The live reproduction shows that the key case is not a corrupted conversation or an expiring local pending action: it is a shared, single-use DSS action that may be resolved through the UI or another client. This smaller follow-up documents the verified recovery behavior without adding a local pending-action API that could be stale relative to DSS.

## Validation

- `uv run pytest tests/test_cobuild.py`
- Focused `pre-commit` checks
- `git diff --check`


pmasiphelps (approved):
looks great - and thank you @mjeckmans for identifying this and helping us resolve :)

PR #113, opened 2026-08-17
docs(cobuild): strengthen pending turn polling guidance

nmartorell:
## Summary
- Clarify that agents must continue long-polling a retained Cobuild turn until it reaches a terminal result.
- Allow independent work while a turn is pending, while requiring a status check before dependent actions or completion.

## Motivation
Long-running Cobuild work remains active after an `in_progress` response, but the previous guidance did not clearly require continued polling. This makes the retained-turn workflow explicit without increasing polling frequency or changing MCP behavior.

This guidance was strengthened after reviewing an observed long-running Cobuild workflow where the agent stopped checking a still-active turn before its terminal result was retrieved.


pmasiphelps (approved):
looks good!

PR #117, opened 2026-08-18
docs: add capabilities documentation

yjouini:
Introduce comprehensive documentation for user-facing capabilities and implement validation tests to ensure consistency between the documentation and registered tools. This update enhances clarity on what Headless can inspect and what Cobuild can build.


nmartorell:
Nice addition @yjouini - although this information is already present in the skills files, I think having one human-intended place with this information is useful (I also really like that the tests keep the list of tools current).

I’d suggest moving this to docs/capabilities.md and linking it from the README, rather than adding another top-level Markdown file. That keeps it findable for users while keeping the repository root focused.

I would keep the capability boundary and tool tables, but remove the ‘Requirements and Installation’ section (it only links to the README) and the ‘Maintenance’ footer (that is contributor/test-maintenance guidance rather than user-facing documentation).

I’d also position the page explicitly as a quick capability boundary/tool matrix, with the skill references remaining the source for the operating workflow. One small clarification: local profile actions (configure_instance, switch_instance, and delete_instance) should be separated from writes to Dataiku objects, since they do not fit the bootstrap/cross-project/admin categories.

yjouini:
Thanks @nmartorell ! Addressed :
- Moved the page to docs/capabilities.md and linked it from the README.
- Removed the installation and maintenance sections.
- Positioned it as a quick capability/tool matrix, with the skill references as the operating guide.
- Separated local profile actions from Dataiku object writes.

nmartorell (approved):
Approved. I made a small README follow-up that moves the capability-matrix link into its own `Capability reference` section after `What it does`, where the Cobuild/direct-action distinction has context. If you disagree with that placement, feel free to revert commit 9730877 and merge this PR as-is.

PR #118, opened 2026-08-19
ci: lock PEP 723 MCP runtime

nmartorell:
## Summary

- Commit a PEP 723 lockfile at `bin/run_mcp.py.lock` and launch the MCP server with `uv run --locked --script` from both Codex and Claude plugin configurations.
- Verify lock validity in PR CI, and add a pre-Commitizen release gate that refreshes the resolution with `--upgrade` and stops before versioning or tagging if permitted transitive dependencies have changed.
- Document the lock lifecycle and cover the required lockfile in the launcher metadata test.

## Motivation

On a cold machine, an unlocked `uv run --script` must resolve the full dependency graph before the MCP server can answer its initialize request. Codex/Claude MCP default startup timeouts (10s) can make that look like an installation failure; restarting after the first installation succeeds only because uv has populated its cache.

The committed lock removes dependency resolution from normal startup and gives every supported platform the same fully pinned direct and transitive graph. `--locked` fails clearly if that graph is missing or stale rather than silently resolving on a user machine.

The release-only `--upgrade` check preserves an intentionally lightweight maintenance model: direct dependencies remain manually pinned, while no new release can be cut if newer compatible transitive packages (including security fixes) are available. The resulting lock change must be reviewed, tested, and committed through a PR before tagging.

## Validation

- `uv lock --script bin/run_mcp.py --check`
- Locked MCP initialize handshake
- `claude plugin validate .`
- `uv run pytest` (212 passed)
- `uv build`


nmartorell:
## Startup timing evidence

I ran end-to-end MCP `initialize` handshakes in disposable copies of the plugin, using fresh temporary uv caches for cold runs and the same cache for warm runs. “Unlocked” means no adjacent script lockfile and `uv run --script`; “locked” means the committed `bin/run_mcp.py.lock` and `uv run --locked --script`.

| Cache state | Locked | Unlocked | Interpretation |
| --- | ---: | ---: | --- |
| Cold: empty uv cache | 20.88 s | 24.42 s | Locking saved 3.54 s by avoiding dependency resolution, but both runs remain over the Codex/Claude 10 s startup deadline because downloads, package preparation, installation, and imports still occur. |
| First reuse after cold provision | 5.93 s | 7.01 s | Locking saved about 1.08 s in this sample. This is the case where the improvement could matter on a slower platform, such as a Windows machine with slower filesystem/endpoint-security overhead. |
| Fully warm: steady-state median | 1.59 s | 1.57 s | No material difference. Python imports and MCP initialization dominate once the environment is warm. |

Additional locked cold-run detail: uv prepared 72 packages in 9.47 s (including downloading artifacts and building the sdist-only `dataiku-api-client`), then installed them into its script environment in 0.59 s. A standalone no-lock resolution took about 3.00 s; locked validation took about 0.08 s.

pmasiphelps (approved):
looks great - getting that startup time down!!

PR #121, opened 2026-08-20
fix: redact encoded connection keys

pmasiphelps:
modifies the `_redact_sentitive_data` function used with the `get_connection_info` tool, now checks if returned key/value pairs contain "accesskey", "passphrase", "privatekey", "secretkey" instead of just checking if these are suffixes. previously, this missed one key: privateKeyB64.


nmartorell (approved):
@pmasiphelps lgtm! My only comment would be to add a note in RELEASE.md mentioning that dip should be re-scanned on dataikuapi update.

pmasiphelps:
@nmartorell makes sense. let's think about scanning dip in a different PR...

PR #123, opened 2026-08-24
feat: add project settings tools

pmasiphelps:
- add two tools: `get_project_settings` and `update_project_settings` 
- skill/ref docs covering flow display, pipelines, Python/R default code envs, and containerized execution defaults. the `update_project_settings` tool **can only update these documented settings**. anything else is rejected. @nmartorell thoughts on this? i leaned conservative...
- removed the unused `deep_merge_dict` util function, replaced with `_apply_json_merge_patch` directly in projects.py, with added capability to remove fields where you pass `null` to it 

@pportal2 this should eliminate the need for separate migrate tooling for flow display settings edits


dataiku_mcp/tools/projects.py
```diff
@@ -136,6 +218,49 @@ async def get_project_variables(project_key: str, ctx: Context) -> str:
...
+
+    def _run():
+        raw = get_dss_client().get_project(project_key).get_settings().get_raw()
+        return raw["settings"]
+
+    return compact_json(await run_blocking(_run))
+
+
+@mcp.tool()
+async def update_project_settings(
```
yjouini, line 257: Hey @pmasiphelps I think that code environment updates can leave contradictory settings, for example, mode="EXPLICIT_ENV" while useBuiltinEnv=true. 

the API updates these fields together for both Python and R (https://github.com/dataiku/dataiku-api-client-python/blob/de83b58cc2d5e1b539fd44797778d56e20f1d367/dataikuapi/dss/project.py#L3566-L3591))

We should enforce mode-dependent invariants, what do you think ?

yjouini (changes requested):

dataiku_mcp/tools/projects.py
pmasiphelps: great catch @yjouini ! changed it :)

yjouini (approved):
LGTM !

PR #91, opened 2026-08-06
feat(plugins): ship Agent Plugins v1.0.0 portable package

crmapj:
## Summary

Adds portable [Agent Plugins](https://agent-plugins.org/) v1.0.0 packaging so any conforming client can load this repo as a plugin without harness-specific manifests.

- **Portable floor:** root `plugin.json` + root `mcp.json` (stdio `dataiku` → `sh ${PLUGIN_ROOT}/bin/launcher.sh`) + existing `skills/`
- **Additive migration:** keep `.claude-plugin/`, `.codex-plugin/`, and `.mcp.json` as parallel legacy packaging (not reverse-domain Agent Plugins extensions)
- **Launcher:** prefer standard `PLUGIN_ROOT` / `PLUGIN_DATA`, fall back to Claude vars, then local checkout `.deps/`
- **Release lockstep:** include portable `plugin.json` in Commitizen `version_files` (version-key selector so schema `1.0.0` URLs are not rewritten) and bump.yml verification
- **Release atomicity:** bump with `push: false`, verify manifests + `$schema` URLs, then push commit/tags

## How we ship plugins today vs after this PR

| Surface | Before | After |
| --- | --- | --- |
| Claude Code | `.claude-plugin/` | unchanged + portable core available |
| Codex | `.codex-plugin/` | unchanged + portable core available |
| Contributor dogfood | `.mcp.json` | unchanged |
| Agent Plugins clients | n/a | root `plugin.json` + `mcp.json` + `skills/` |

### Adversarial review — GPT-5.6 Sol @ xhigh (cbrain job `414bbe8c-…`)

**Verdict on first pass: REQUEST CHANGES.** Findings addressed in the follow-up commit:

| Severity | Finding | Resolution |
| --- | --- | --- |
| Blocker | `tomllib` import breaks Python 3.10 CI collection | Use `importlib.metadata.version` instead |
| High | Manifest verify ran *after* remote push/tag | `push: false` → verify → explicit push |
| Medium | Happy-path-only tests | Schema-selector sim, stricter cwd, local fallback, launcher path existence |
| Medium | AGENTS.md stale | Updated |
| Medium / residual | Credentials still under `~/.dataiku`, not `PLUGIN_DATA` | **Accepted residual** — changing config location is a separate behavior change; runtime/cache already use `PLUGIN_DATA` |
| Residual | Windows/`sh` | Documented POSIX assumption for portable MCP |

[...]

PR #129, opened 2026-08-26
feat(plugins): add agent-assisted setup

JedIV:
## Summary

- add an explicit `setup` skill that checks `uv` first, offers the official platform installer with explicit approval, and verifies the Dataiku connection
- make `Set up Dataiku Headless` the primary Codex starter prompt
- route first-install and MCP recovery from the main Dataiku skill into the setup workflow
- update the README so users can install the plugin before handling the runtime prerequisite

This complements #91, which adds portable Agent Plugins packaging for Cursor and other conforming clients.

## Test plan

- [x] plugin-creator validation
- [x] skill validation for `setup` and `dataiku-headless`
- [x] `claude plugin validate .`
- [x] `uv run pre-commit run --all-files --show-diff-on-failure`
- [x] `uv run pytest` (235 passed)
- [x] `uv build`
- [x] `uv run cz check --rev-range origin/main..HEAD`


pmasiphelps (approved):
changed the name of the new skill from setup -> dataiku-headless-setup so that it wont conflict with other skills. looks good!

PR #130, opened 2026-08-26
chore(deps): refresh MCP script lock

pmasiphelps:
Refreshes bin/run_mcp.py.lock so the release workflow's freshness check passes.\n\nValidation:\n- uv run pre-commit run --all-files --show-diff-on-failure\n- uv run pytest\n- uv build


nmartorell (approved):
👍

PR #133, opened 2026-08-26
feat(cobuild): return the project UI URL in Cobuild payloads

### LINKED ISSUE #127: feat: return the Cobuild conversation URL in Cobuild tool payloads
## Problem

Nothing in the Cobuild tool surface returns a link to the Cobuild conversation in the Dataiku UI. The operator is told a `conversation_id` and a `project_key`, and has to find the conversation in the product themselves.

Raised at the 24 Aug enablement session (Headless Chat Q&A, Amey Ambade): "Generally, when are we given an indication of Headless triggering a Cobuild chat?"

## Suggested fix

Derive the conversation URL from the active instance URL plus project key and conversation id, and include it in:

- `start_cobuild_conversation` (`cobuild.py:216`)
- `_terminal_turn_result` (`cobuild.py:64`) — especially when `is_confirmation_request` or `is_question_request` is set
- `_pending_turn_result` (`cobuild.py:108`) and `_turn_status_payload` (`cobuild.py:124`)
- `list_cobuild_conversations` (`cobuild.py:381`)

Then update `skills/dataiku-headless/references/cobuild.md` to tell the agent to surface the link when a turn is waiting on a human decision or is still running after a poll.

yjouini:
Closes #127

## What changed

Every Cobuild payload now carries a `project_url` pointing at the Dataiku UI, so an operator can open the project Headless is driving instead of hunting for it — especially when a turn is parked on a human decision.

Payload builders covered:

- `start_cobuild_conversation`
- `_terminal_turn_result` — including the `is_confirmation_request` / `is_question_request` cases
- `_pending_turn_result` (which also gained `project_key`)
- the SDK-exception turn result in `_start_turn`
- `list_cobuild_conversations` (new `project_url` column)

`_turn_status_payload` is deliberately unchanged: its output is embedded in error text as the literal argument object to pass to `get_cobuild_turn_status`, so adding a non-parameter key there would invite the agent to send a bogus argument. Instead, the two `_require_turn_available` messages that embed that payload — the "turn is queued/in progress" and "terminal result not observed" cases — now append a sentence with the URL, which is the operator-visible surface the ticket was after.

**No conversation-scoped DSS front-end route could be verified, so this is the project-scoped URL — not a conversation deep link.** Evidence:


pmasiphelps (approved):
really nice, friendly PR

PR #135, opened 2026-08-26
feat(instances): report the active instance DSS version

### LINKED ISSUE #126: feat: report the instance DSS version and use it in capability guidance
## Problem

Headless has no awareness of the DSS version of the instance it is connected to. `dataiku_mcp/tools/instances.py` exposes name, URL, description and active flag, and nothing reports a version; 

Raised at the 24 Aug enablement session (Headless Chat Q&A, Lennard Henjes): "what happens if I try to do v15 multi-target regression but my instance is on 14.7". Clement Stenac confirmed: "No, it doesn't."

## Impact

When an operator asks for a capability their instance does not have, Headless discovers the gap only when the SDK call or Cobuild turn fails, then improvises a substitute or stops. The user gets no indication that the request was valid but the instance is too old. Two reports in #ask-dataiku-headless are this failure:

- Venkat Ravuri, 20 Aug: "This DSS version does not expose any visual GenAI recipe types (Prompt, LLM classification, or Summarization); only standard visual recipes are available. I therefore cannot create a compliant visual substitute"
- Dobrina Ivanova, 14 Aug: asked for a semantic model, Cobuild returned manual setup instructions, Headless reported it "cannot create or configure them"

## Suggested fix

1. Capture the instance DSS version when the client is built and include it in `get_current_instance` and `list_instances` output.
2. Have the skill state the version in its instance summary, so it is in context before any build request is planned.
3. When a requested capability is unavailable, distinguish "not available on this DSS version" from "not supported by Headless" in the message, and name the version that would provide it.

## Evidence

- `dataiku_mcp/tools/instances.py:16` (`list_instances`) and `:66` (`get_current_instance`) — no version field
- `dataiku_mcp/tools/utils/auth.py:47` (`get_dss_client`) — natural capture point
- Source: Headless Chat Q&A & Summary, 24 Aug 2026 enablement session

yjouini:
Closes #126

## Behavior changed

- `get_current_instance` now returns `dss_version`, the DSS version of the connected instance.
- `list_instances` gains a `dss_version` column, populated for the **active instance only** (empty for the others).
- `skills/dataiku-headless/SKILL.md`: the instance rule now tells the agent to read `dss_version` before planning a build, and a new shared rule requires a capability gap to be reported as either "not available on this DSS version" (naming the instance version and the version that introduces the feature) or "not supported by Headless".
- `skills/dataiku-headless/references/cobuild.md`: a Cobuild refusal, manual-setup answer, or "asset type unavailable" answer must be checked against `dss_version` before being called a Headless limitation, and a version-gated asset type must not be silently substituted.
- `docs/capabilities.md`: instance row updated.

## Scope decision

Versioning every configured instance would mean one HTTP round trip per configured instance on every `list_instances` call — a latency regression on a tool that was previously config-only. `list_instances` therefore makes at most one call, for the active instance, and its docstring says so.


THE-MOLECULAR-MAN:
This would be helpful for something I'm working on as well.

pmasiphelps (approved):
@yjouini looks good! i made a few changes. what do you think?

- changed "DSS" to "Dataiku" 
- removed the instance version from list_instances. since this only returned the version for the active instance, i think it's simpler to just include it in get_current_instance
- modified get_current_instance to not fail on the instance version API call if the instance can't be reached, or if the API key is invalid. instead, returns ""
- removed some tests - just 2 now

yjouini:
Thanks Pat ! these changes look good to me.

PR #136, opened 2026-08-26
feat(setup): warm the uv runtime before first server launch

yjouini:
## Summary

Adds a runtime warm-up step to the `dataiku-headless-setup` skill added in #129. The host gives up on an MCP server that is still downloading dependencies on its first launch, which is the most common cause of a plugin that appears installed but never becomes available — the "restart Codex / still initializing" reports in #ask-dataiku-headless.

The new step 5 runs the server once with stdin closed so it exits immediately, populating the uv cache and the script environment before the host ever launches it. Subsequent steps are renumbered; the frontmatter description now mentions the warm-up.

## Why run-and-exit rather than `uv sync`

`uv sync --locked --script` looks like the natural fit but only gets partway. Measured on this branch with a throwaway `UV_CACHE_DIR` (so the real cache was untouched), timing a full MCP `initialize` round-trip against `bin/run_mcp.py`:

`uv sync --script` caches the downloads but leaves the environment to be created on first launch, so the host still sees a slow first start. The skill therefore uses the run-and-exit form and says explicitly not to substitute sync. A cold cache for this script is ~148 MB, so on a slow or proxied network the 6.36s figure is a floor, not a typical case.


pmasiphelps (approved):
looks good! i made a slight change to the wording of the run_mcp relative path....codex complained about the previous wording. totally fine if you don't like the new wording. approved either way.

PR #138, opened 2026-08-27
feat(dss-plugins): manage installed instance plugins

yjouini:
## Summary

Adds three MCP tools for installed Dataiku instance plugins: `list_plugins`, `update_plugin`, and `delete_plugin`.

- Find installed plugins by id, label, or tags, with optional metadata and the bound code-environment name. Plugin configuration, presets, and parameter sets are never returned.
- Update an installed plugin from the store or a local directory/ZIP. Local uploads validate the manifest id and flatten a single wrapper directory.
- Delete a plugin after checking usages; report refusals, asynchronous progress, and reload/restart requirements explicitly.

Plugin installation and initial code-environment creation/binding are out of scope. A missing bound environment does not establish whether a plugin needs one. The skill routes initial setup to the Dataiku UI pending a supported discovery and setup workflow with R&D.

`update_plugin(rebuild_code_env=true)` remains available for an already-bound environment. It updates the plugin and then rebuilds that environment; it does not create or bind one. If none is bound, the rebuild is reported as skipped with manual setup guidance. Failed or unfinished rebuilds are reported separately from the plugin update.

## Outcome handling

Plugin actions can return HTTP 200 with `success: false`; the tools check the reported outcome. Code-environment rebuilds use nested `error`/`fatal` reports, since `messages.success` alone is not a reliable success flag. Inline future results and ongoing operations are distinguished, and running futures remain followable with `get_future_status`.

Deletion checks usages and unresolved components before proceeding. `force=true` bypasses the usage safeguard only when explicitly requested. Deleting a plugin does not delete its code environment.

## Supporting changes

Updates the tool inventory, capability matrix, README, and administration references. Shared timeout validation and error-message extraction live in the existing utilities; jobs and scenarios reuse the timeout validator, including `abort_job`. Test clock helpers are shared in `tests/utils/fakes.py`.

## Validation

- `uv sync --locked`
- `uv run pytest`: 291 passed on Python 3.14
- `uv build`: wheel and source distribution built
- Built-wheel import smoke test on Python 3.12 outside the checkout
- Locked stdio launcher initialization smoke test
- Python syntax checks and skill validation

No live Dataiku mutations were performed for this revision.


pmasiphelps:
@yjouini my main concern is that an agent can't know that a plugin requires a code env initially, when no code env is bounded to it. this appears to be a dataiku api gap, im not sure there's anything we can do about it. do you think we can keep just the list_plugins, update_plugin, and delete_plugin tools and associated skill references - then work with r&d on how to best handle plugin code envs?

nmartorell:
The narrowed scope resolves the initial setup ambiguity: an unbound plugin does not establish whether a code environment is required, so initial creation and binding belong in the Dataiku UI. One recovery path remains unclear for an already-bound environment. `update_plugin(rebuild_code_env=true)` can time out before its rebuild starts; after following that update to completion, the documented retry is another `update_plugin(..., rebuild_code_env=true)`, which repeats the plugin update rather than rebuilding alone. Please clarify the intended supported recovery: provide a rebuild-only path for an already-bound environment, or explicitly route this retry to the Dataiku UI instead of advising a repeated update.

yjouini:
Updated the recovery path: if the update outlasts the wait budget, ask the user to rebuild the existing environment in the DSS UI

nmartorell:
I added two small, independent follow-up commits: fix(plugins): reject symlinks in local plugin directories prevents a local-path update archive from including files outside the plugin tree; feat(plugins): paginate plugin deletion blockers adds a read-only paginated inspection path so an agent can review the full delete impact before using force. Both are intentionally isolated and can be reverted independently if you prefer not to take either approach.

pmasiphelps (approved):
looks great

PR #139, opened 2026-08-28
feat(cobuild): add tool titles, intent descriptions, and annotations

yjouini:
## What

Catalog-time metadata for the six Cobuild tools. No tool is renamed, no argument or default changes, and every result payload is byte-for-byte what it was.

- **Titles** — hosts show `Send Cobuild Message`, not `send_cobuild_message`.
- **Descriptions** — one line per tool, saying what it is for and when to reach for it.
- **Parameter descriptions** only where a name and type cannot convey correct use: `project_key` (must match the conversation), `allow_edit_project`, `answers` (empty when rejected), `rejected`, `used_custom_answer`. `choice` becomes `Literal["APPROVE", "CANCEL"]`, so the allowed values ride in the type and reach the schema as an enum; the boundary check still covers direct callers.
- **Annotations classified from each handler's code paths**, not from its name:

  | Tool | `readOnlyHint` | `destructiveHint` | `idempotentHint` | `openWorldHint` |
  | --- | --- | --- | --- | --- |
  | `start_cobuild_conversation` | false | false | false | false |
  | `send_cobuild_message` | false | true | false | false |
  | `answer_cobuild_confirmation` | false | true | false | false |
  | `answer_cobuild_question` | false | true | false | false |
  | `get_cobuild_turn_status` | true | false | omitted | false |
  | `list_cobuild_conversations` | true | false | omitted | false |

## Why these classifications

Editing is dynamic through `allow_edit_project`, but annotations are static, so the four write-capable tools are catalogued at their maximum possible effect. `answer_cobuild_question` counts because an answer can resume an edit-enabled build. Polling and listing observe only — the internal `observed` bookkeeping records that a caller saw a retained result; it changes no Dataiku state and no later poll's result. `openWorldHint: false` throughout: every call is bounded to the configured instance and a named project.

An annotation is a catalog-time label, never an authorization. It does not replace `allow_edit_project`, the exact-turn check, or Dataiku permissions.

## Catalog cost

The catalog is re-sent on every turn of every session, used or not, so a description carries what a caller needs to pick the tool and nothing more. Workflow detail and safety rules stay in `skills/dataiku-headless/references/cobuild.md`, which loads on demand and is unchanged here.

Descriptions plus input schemas for the six tools go **1,726 → 2,258 chars** (~+130 tokens). The convention is written down under **Tool Metadata** in `CODING_STANDARDS_AND_STRUCTURE.md`, with a pointer from `AGENTS.md`, before the same density reaches the other 117 tools.

## Docs

The two `docs/capabilities.md` rows whose intent was narrower than the tool behavior (`send_cobuild_message` also inspects; `answer_cobuild_confirmation` is specifically a deletion), plus the row for `answer_cobuild_question`. No README or tool-count change — the registered surface is identical.

## Scope

Metadata only, deliberately. Structured outputs (real `outputSchema` plus `structuredContent`) are a separate concern. No new tests: the change is declarative and the existing suites — tool surface, capabilities doc, Cobuild behavior — cover it unchanged.

## Verification

- `uv sync --locked`, `uv run pre-commit run --all-files`, `uv build` — clean
- `uv run pytest` — 237 passed
- `uv run cz check --rev-range origin/main..HEAD` — clean
- Drove the pinned plugin runtime (`uv run --quiet --locked --script ./bin/run_mcp.py`) over stdio and confirmed titles, annotations, and parameter descriptions reach the wire; all 123 tools keep the shape they had before


pmasiphelps (approved):
looks great. no issues. really nice add :)

PR #140, opened 2026-09-02
feat(tools): execution overrides tool

yjouini:
## Summary
- new tool`set_container_exec_config` taking an `object_type`
- support `INHERIT`, `NONE`, and `EXPLICIT_CONTAINER` on recipes, ML tasks, saved-model retrain, WebApp backends, Knowledge Banks, and agent tools
- refuse a recipe that exposes a container selection in more than one location rather than writing to one Dataiku may not read
- document the exception once as placement-only, and route to it from a new `references/container-execution.md`

| `object_type` | `object_id` | Settings path written |
|---|---|---|
| `recipe` | recipe name | `params.containerSelection`, `params.engineParams`, or the visual payload's `engineParams` |
| `ml_task` | analysis id (single-task analysis) | mltask `containerSelection` |
| `saved_model` | saved model id | `miniTask.containerSelection` (retrain placement) |
| `webapp` | WebApp id | `params.infra.containerSelection` (backend placement) |
| `knowledge_bank` | Knowledge Bank id | `containerExecSelection` |
| `agent_tool` | agent tool id | `params.containerExecSelection` |

Objects that run no code of their own — an MLflow-imported saved model, a built-in agent tool — expose no selection, and the tool fails without saving rather than inventing one. Same for an analysis that does not hold exactly one ML task.

The exception stays placement-only: it does not open Knowledge Banks or agent tools to any other direct write. Agents and Retrieval-Augmented LLMs are deliberately out of scope — their placement sits per version under a flavor-specific settings block, so a write needs a rule for which version and which flavor is live.


pmasiphelps (approved):
@yjouini this looks great. i made two changes - what do you think?
1. Removed direct saved-model container edit - I couldn't find in the UI where this param can be changed! I think the real thing to change is the train recipe before the saved model - so I removed saved model as an allowed object type for container conf changes, and modified the skills to instead use the prev train recipe. 
2. (super minor) Narrowed recipe payload error handling to preserve real SDK failures
Approved and you can merge if you like the changes!

yjouini:
LGTM !

PR #143, opened 2026-09-04
fix(onboarding): handle "discovered" but bad credentials better

pmasiphelps:
When helping two customers install headless, I saw headless "find" some credentials (i believe in .dataiku directory), therefore skip the configure_instance tool call in the setup skill steps. this PR adds checks that get_current_instance actually works on the "found" instance credentials, and if it doesnt, continues to configure_instance.

Also, /instance-info is not actually admin permission-gated, so removed that language


nmartorell (changes requested):
This is great.
My one concern is that I don't think get_dataiku_version should live in auth.py. (I know this PR doesn't introduce this method, but we should probbaly just move it inline into the get_current_instance call).

pmasiphelps:
@nmartorell fixed! good catch

nmartorell (approved):

PR #144, opened 2026-09-04
feat(jobs): add abort_job tool to stop a running Dataiku job

pportal2:
Adds an abort_job tool next to the existing job tools (run, status, log, wait). Cobuild cannot abort a job, so once an agent starts a slow recipe it must wait for the job to finish before it can change the recipe. This came up in an Alteryx migration where a Snowflake recipe blocked the agent for more than 20 minutes. The tool sends the abort, waits for the terminal state by default, and reports job_aborted, abort_requested, abort_still_pending or job_already_finished.


pmasiphelps:
@pportal2 looks great! i made a couple changes. what do you think?

- Fixed abort-result reporting: a job that finishes as DONE or FAILED **after** an abort request is no longer incorrectly reported as aborted. Added a clear job_finished_after_abort_request outcome with guidance to inspect outputs before rerunning.
- Made post-abort polling failures recoverable: returns abort_poll_failed with the known job identity instead of throwing away the accepted abort outcome.
- Added two focused regression tests for those^ cases.
- Updated the routing guideline to state that abort_job can stop an existing accessible job, rather than implying it only stops jobs started by the current headless direct exception tools.

pportal2:
@pmasiphelps I think it's great! Let's merge this thing

pmasiphelps (approved):

PR #145, opened 2026-09-09
fix(cobuild): add the trailing slash to the project Flow URL

pportal2:
Every Cobuild tool payload carries a project_url built by _project_url in dataiku_mcp/tools/cobuild.py. It returned /projects/<KEY>/flow, which DSS does not route, so the link the agent gives the user does not open. /projects/<KEY>/flow/ works. This appends the slash.

Example: 
https://agent-dev-kit-testing.se-platform.dataiku-sandbox.io/projects/PROJ_E26CD6_82b0_FRAUDOPE/flow
vs
https://agent-dev-kit-testing.se-platform.dataiku-sandbox.io/projects/PROJ_E26CD6_82b0_FRAUDOPE/flow/


pmasiphelps (approved):
lol thank you :)

PR #146, opened 2026-09-09
feat(tools): add titles, intent descriptions, and annotations to inspection tools

yjouini:
Part 1 of 5. Extends #139's tool metadata to 15 read-only modules: title, one-line description, annotations, and parameter descriptions only where a name and type don't suffice.

All read-only, so all carry `readOnlyHint: true` and omit `idempotentHint`.

Dropped the `Args:` blocks on `get_licensing_status`, `list_container_exec_configs`, and `list_spark_configs`, which were duplicated in `references/administration/`. Kept their admin precondition as an `Admin only.` suffix, since it decides whether the tool is usable at all.

The standards doc gains the three rules this rollout applied: when `destructiveHint` is true, why `openWorldHint` is false throughout, and when a parameter should become a `Literal`.


pmasiphelps (approved):
lgtm

PR #147, opened 2026-09-09
feat(tools): add titles, intent descriptions, and annotations to data tools

yjouini:
Part 2 of 5. Same treatment for datasets, flow, connections, data quality, libraries, managed folders, evaluation stores, and LLMs.

Destructive: `export_dataset` and `write_project_library_file` replace a file when `overwrite` is set; `upload_file_to_managed_folder` replaces whatever is at its target path. `create_upload_dataset` is not — it raises on an existing name rather than replacing.

Three `Literal` conversions (evaluation store `flavor`, flow `object_type`, connection `connection_category`) drop the `require_allowed_value` they replace. Three parameters stay plain strings on purpose: `search_type`, `purpose`, and `source` normalize case, so a `Literal` would reject values that work today; `connection_type` has 63 values, which costs more in every catalog send than a description does.

`get_data_quality_rule_history.rule_ids` had no type at all in the advertised schema. Now annotated with the union its parser already accepts, so nothing is narrowed.

Worth knowing: dropping `require_allowed_value` moves an invalid value from a project `ValueError` with an `Allowed values` message to a Pydantic error at the boundary. Same gating, different text. Nothing in the repo matches on it.


pmasiphelps (approved):
lgtm

PR #148, opened 2026-09-09
feat(tools): add titles, intent descriptions, and annotations to admin tools

yjouini:
Part 3 of 5, and the only one with behavioral changes. Split out so review lands here.

**Ten `Literal` conversions** (`source_type` ×5, code-env `language` ×4, `search_mode`, `python_interpreter`) drop the `require_allowed_value` they replace. `require_allowed_value` does no normalization, so the types are exact equivalents; what changes is that an invalid value now raises a Pydantic error at the boundary instead of a project `ValueError` carrying an `Allowed values` message. Nothing in the repo matches on that text.

**`utils/identity_sources.py` now owns a shared `IdentitySourceType` type** instead of a runtime validator, since every call site was a tool parameter. Six enum constants left dead by the conversions are removed.

**Four latent bugs.** Removing the validators left dangling `if` guards that had wrapped them, each then governing an unrelated statement. The worst: in `list_groups`, `offset` validation ended up inside `if source_type is not None:`, so `offset` went unvalidated whenever `source_type` was omitted. Same shape in `update_group` (the `mappings` dict) and twice in `code_environments.py`. All fixed.

**One test retargeted.** `test_create_code_env_rejects_unknown_python_interpreter` called the handler directly, bypassing the boundary the check now lives at, so it asserts the advertised enum instead. It reads that enum recursively — Pydantic nests an optional annotated `Literal` one `anyOf` deeper on 3.10 than on 3.12+, which is what broke CI here initially. Verified on 3.10, 3.12, and 3.14.

The ~35 permission booleans on `create_group`/`update_group` are left undescribed: their names say what they mean, and `references/administration/groups.md` documents which imply others.


pmasiphelps (approved):
lgtm

PR #149, opened 2026-09-09
feat(tools): add titles, intent descriptions, and annotations to execution tools

yjouini:
Part 4 of 5. Same treatment for jobs, scenarios, projects, and instances.

Execution tools are catalogued at their maximum possible effect, so anything that runs existing Dataiku logic is destructive: `build_datasets`, `run_recipe`, `run_scenario` (rewrite their outputs; a scenario can do whatever its steps allow), and `abort_job` (leaves partial output in place, idempotent since a terminal job isn't re-aborted). Also `update_project_settings` (nulls remove fields) and `set_project_variables` (replaces the object wholesale, dropping anything omitted). `delete_instance` forgets local config only; the instance is untouched. The five polling tools are read-only.

`job_type` becomes a `Literal`, dropping the `require_allowed_value` it replaces — same error-text caveat as #147. `test_build_datasets_invalid_job_type_rejected` called the handler directly, bypassing that boundary, so it asserts the advertised enum instead.

**Rebase note, worth a look from whoever reviewed #143:** #143 gave `get_current_instance` a `connection_status` field and a multi-line docstring, which conflicted with this PR's trim. I folded the new signal into the one-line form ("Confirm which instance is active, whether it can be reached, and its version") since the `connection_status` semantics are already in `dataiku-headless-setup/SKILL.md` step 8. Worth a look.


pmasiphelps (approved):
lgtm

PR #150, opened 2026-09-09
docs(agents): record that every tool now carries advertised metadata

yjouini:
Part 5 of 5. **Merge last.**

#139 noted that existing tools predated the metadata rules and were being migrated module by module. #146, #147, #148, and #149 finish that, so the note becomes a rule.

One line. Split out from #149 rather than bundled, because the claim is only true once all four have merged and merge order isn't something a bundled change can guarantee.


pmasiphelps (approved):
lgtm

